PR #217, opened 2015-02-02
Add sync info to machines table

robkooper:
See https://github.com/PecanProject/pecan/issues/149

### DIFF
--- a/app/views/machines/_index_table.html.erb
+++ b/app/views/machines/_index_table.html.erb
@@ -6,6 +6,7 @@
     <thead>
       <tr role="row">
         <th class="sorting_asc" role="columnheader" aria-controls="DataTables_Table_0" style="width: 40px"><%= sortable "machines.hostname", "Hostname" %></th>
+        <th role="columnheader" aria-controls="DataTables_Table_0" style="width: 40px"><%= sortable "machines.sync_host_id", "Sync Host ID" %></th>
         <th role="columnheader" aria-controls="DataTables_Table_0" style="width: 162px">Actions</th>
       </tr>
     </thead>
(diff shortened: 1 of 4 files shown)


gsrohde:
Rob: I made some changes -- partly better integrity constraints, also avoiding a table that is mostly NULLs.  The basic Rails functionality is there, but still some kinks perhaps to work out.

If you hate this, we can roll back to your version or maybe come up with some kind of compromise.

robkooper:
Not sure what the problem is of having null columns in the table, I think it makes it harder to read to have a second table that has some of the values stored. Otherwise it would probably be better to have a special sync table that has the fields needed and an ID to the machine that is hosting the data. Then have a special edit section that will allow a person to add/remove a row to this sync table.

gsrohde:
I'm not exactly sure what your objection is, but I now see that using an SQL view was a somewhat awkward attempt to avoid as much as possible having to change your Rails code, and I propose doing one of the following:

1. Dispense with the SQL `machines` view and adapt the Rails code to work directly with the `database_machines` table, adapting the Rails view (and model and control) as necessary.
2. Using two totally separate tables, one for file-storing machines and one for database-sync machines.

If we go with option 1, I would change the foreign key in the `database_machines` table to refer to `machine_list` by id rather than by hostname to be in conformity with all the other foreign keys in the database.

With either option, I leave open the question of whether to have the two tables share one set of Rails views, which would interact with either or both of the tables as necessary, or whether to have two separate sets of views.

Also, we can keep the name "machines" for the existing table (instead of renaming it to "machine_list") and call the new table anything you like--"database_machines" or "sync_machines" or whatever.

In any case, I still think it a good idea to use a single column of range type to store the sync range.

PR #218, opened 2015-02-04
add IDs to many-many tables

robkooper:
This ID will be used to find those tables that should be synchronized
and what data was entered at what site.

The id's are set to auto increment and will be defaulted to be in the
range of EBI.

The script fix.many.id.sh can be used to fix the auto increment ID to
be in the right range and will try to set the ID's of those that are
entered at the remote site (this is done by looking at the id's of the
elements referenced and see if any of them fall inside the range of the
site). To run the scrip use the following syntax: `MYSITE=1
./fix.many.id.sh`. The arguments are the same as load.bety.sh in pean.

@mdietze Currently this script will only need to be executed at BU and
should not be run at EBI.

### DIFF
--- a/db/migrate/20150202215147_add_id_many_to_many.rb
+++ b/db/migrate/20150202215147_add_id_many_to_many.rb
@@ -1,14 +1,14 @@
 class AddIdManyToMany < ActiveRecord::Migration
   def self.up
     execute %{
-      ALTER TABLE "citations_sites" ADD "id" serial;
-      ALTER TABLE "citations_treatments" ADD "id" serial;
-      ALTER TABLE "inputs_runs" ADD "id" serial;
-      ALTER TABLE "inputs_variables" ADD "id" serial;
-      ALTER TABLE "managements_treatments" ADD "id" serial;
-      ALTER TABLE "pfts_priors" ADD "id" serial;
-      ALTER TABLE "pfts_species" ADD "id" serial;
-      ALTER TABLE "posteriors_ensembles" ADD "id" serial;
+      ALTER TABLE "citations_sites" ADD "id" bigserial;
+      ALTER TABLE "citations_treatments" ADD "id" bigserial;
+      ALTER TABLE "inputs_runs" ADD "id" bigserial;
+      ALTER TABLE "inputs_variables" ADD "id" bigserial;
+      ALTER TABLE "managements_treatments" ADD "id" bigserial;
+      ALTER TABLE "pfts_priors" ADD "id" bigserial;
+      ALTER TABLE "pfts_species" ADD "id" bigserial;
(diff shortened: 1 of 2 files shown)


robkooper:
Fixes https://github.com/PecanProject/pecan/issues/325

db/migrate/20150202215147_add_id_many_to_many.rb
```diff
@@ -0,0 +1,25 @@
...
+    execute %{
+      ALTER TABLE "citations_sites" ADD "id" serial;
+      ALTER TABLE "citations_treatments" ADD "id" serial;
+      ALTER TABLE "inputs_runs" ADD "id" serial;
+      ALTER TABLE "inputs_variables" ADD "id" serial;
+      ALTER TABLE "managements_treatments" ADD "id" serial;
+      ALTER TABLE "pfts_priors" ADD "id" serial;
+      ALTER TABLE "pfts_species" ADD "id" serial;
+      ALTER TABLE "posteriors_ensembles" ADD "id" serial;
+    }
```
gsrohde: Shouldn't these all be "bigserial"?

robkooper: correct, keep forgetting.

gsrohde:
After pulling this, it occurred to me that it would make more sense for the new column to be "sync_host_id".  Much simpler--no autonumbering.  Is there something I'm missing?

robkooper:
auto numbering is easy since it fits in the existing scheme where we assign numbers automatic to id's. Yes we can use the sync_id number everywhere as well but then we have to configure the app to set this id. Using the auto numbers it is automatic.

gsrohde:
@robkooper Can I move the script ./fix.many.id.sh into the scripts directory?  I'd rather not pollute the RAILS ROOT directory.

gsrohde:
@robkooper Also, is there some documentation about running this script that I can point users to in the release notes?

db/migrate/20150202215147_add_id_many_to_many.rb
```diff
@@ -0,0 +1,25 @@
+class AddIdManyToMany < ActiveRecord::Migration
+  def self.up
+    execute %{
+      ALTER TABLE "citations_sites" ADD "id" bigserial;
+      ALTER TABLE "citations_treatments" ADD "id" bigserial;
+      ALTER TABLE "inputs_runs" ADD "id" bigserial;
+      ALTER TABLE "inputs_variables" ADD "id" bigserial;
+      ALTER TABLE "managements_treatments" ADD "id" bigserial;
+      ALTER TABLE "pfts_priors" ADD "id" bigserial;
```
dlebauer: @robkooper do these need to be bigserial? ESRI doesn't like it https://github.com/terraref/reference-data/issues/81 but there may be other options

PR #219, opened 2015-02-05
Adding draft migration for first batch of uniqueness constraints.

gsrohde:
Rob: Here's my first batch of uniqueness constraints.  Let me know if you foresee problems.  Send back with comments or approval.

* [ ] @robkooper please confirm that @gsrohde can deploy this
* [ ] PecanProject/pecan#376 (to use mimetypes) should be merged when this migration is deployed


robkooper:
Why is there the following constraint? spaces are fine in filenames:
```
ALTER TABLE dbfiles ADD CONSTRAINT no_space_in_file_name CHECK (file_name !~ '\s');
```

gsrohde:
If there is no file_name, the column value should be the empty string ('').

robkooper:
I think the constraint of the file_path is to strict. The only invalid chars in a folder name are / and \0. One option to use is:
```
ALTER TABLE dbfiles ADD CONSTRAINT file_path_sanity_check CHECK (file_path ~ '^(([a-z.-]+:)?//?([\.A-Za-z0-9_\-=]+/)+[\.A-Za-z0-9_\-=]*/?)?$');
```
or we might just say that all paths need to be absolute, and thus start with a /, ie:
```
ALTER TABLE dbfiles ADD CONSTRAINT file_path_sanity_check CHECK (file_path ~ '^/');
```

robkooper:
names of pfts can have a whitespace.

gsrohde:
Re pfts.name: can we at least require no leading or trailing space, and no internal space other than bona fide space characters (i.e., no tabs, newlines, non-breaking spaces, carriage returns, form-feeds, etc.)?

robkooper:
Fair enough

gsrohde:
Regarding dbfiles.file_path: The first constraint you put in your comment is even stricter than the one I have!

If the files are always on ebi-cluster.igb.uiuc.edu, then clearly that part of the path is redundant and should be eliminated and we can standardize on absolute paths starting with '/'.  A simpler but slightly more restrictive constraint than your second option (`CHECK (file_path ~ '^/')`) would be `CHECK (file_path ~ '^/[\w ./:-]*$')`--that is, file paths must begin with '/' and consist only of alpha-numeric characters (which includes underscore and some (all?) accented and non-latin letters), periods, hyphens, forward slashes, colons, and spaces.  (Or maybe we shouldn't allow spaces and/or colons.)  Or we could allow only printable ASCII characters with `CHECK (file_path ~ '^/[\x20-\x7e]*$')`.


When you said "The only invalid chars in a folder name are / and \0" did you mean file name?  Even if those are the only characters not allowed by the operating system, we might want to be stricter.  For example, spaces are allowed but can be a pain.  At a minimum, can we restrict to printable ASCII characters?

gsrohde:
Rob, let me know when you are done with this so I can pull it or change it to your specifications.

gsrohde:
@robkooper I updated this a little bit (see commit comment).  The main change is the addition of constraints on the runs table, which I had in a draft of this migration but had not committed.

If these are too much of a pain to deal with right now, I can take them back out and we can maybe add them later.

Issue #41, opened 2014-04-29
Drop deprecated tables and columns


mdietze:
## update: The following tables / fields will be dropped December 2014:

Please feel free to modify this directly. Linked issues should be closed / updated as appropriate
- **tables:** 
  - counties, locations_yields, [2050](https://ebi-forecast.igb.illinois.edu/redmine/issues/2050) - migrated to #177
  - sessions tables [2048](https://ebi-forecast.igb.illinois.edu/redmine/issues/2048) - migrated to #177
- **fields** 
  - workflows.site_id, 
  - workflows.model_id, [2049](https://ebi-forecast.igb.illinois.edu/redmine/issues/2049); 
  - inputs.file_id (#36), 
  - formats.mime_type (replace with formats.mimetype_id)
  - formats.format
### original question

> There are columns (and possible whole tables) that are "left over" from initial designs that were later replaced by different functionality. In lieu of just deleting them immediately (which may break things) is there a way to flag these as deprecated so as to give users a warning that they're likely to go away? It would be good to block new additions to these columns as well. The example I just noticed is the Data Format column in the Formats table, which was replaced by the formats_variables tables a good while back, but it appears that new records using this column have been added since then.

PR #223, opened 2015-02-09
Draft migration addressing issues in GH #41.

gsrohde:
This implements the changes outline in the spreadsheet at https://docs.google.com/spreadsheets/d/1fJgaOSR0egq5azYPCP0VRIWw1AazND0OCduyjONH9Wk/edit?pli=1#gid=1304822510 (see also GH #41 and related issues).

@dlebauer, @robkooper, @mdietze Please comment and suggest modifications if needed.

Marking `formats.mimetype_id` and `yields.entity_id` as foreign keys and imposing value constraints on the new data fields will probably be done in later migrations.


gsrohde:
@dlebauer, @robkooper, @mdietze Please comment and suggest modifications if needed.

db/migrate/20150209174734_add_drop_tables_and_columns.rb
```diff
@@ -0,0 +1,93 @@
+class AddDropTablesAndColumns < ActiveRecord::Migration
```
robkooper: Can it have a more specific name? Like add_constraint_mimetype?

mdietze:
Looking at the Google Doc, on the Drop/Add page, what Issue is associated with dropping Start_date and End_date from the Runs table? Perhaps I'm being absent minded but I don't recall that discussion and can't remember why we'd want to drop those columns.

dlebauer:
@mdietze https://ebi-forecast.igb.illinois.edu/redmine/issues/459 replaced *_date with *_time

robkooper:
Who will update PEcAn to reflect the change to start_date and end_date in the code? Will there be a second pull request that goes with this issue?

dlebauer:
> Who will update PEcAn to reflect the change to start_date and end_date in the code?

First we should resolve the question of whether there is a reason to make the change. I can't recall who suggested it or why this was suggested though it seems to have come up when Pat was here: in addition to  [Redmine 459](https://ebi-forecast.igb.illinois.edu/redmine/issues/459) it was also discussed in [Redmine 188](https://ebi-forecast.igb.illinois.edu/redmine/issues/188).

mdietze:
Ok, that's bizarre, why would Pat have added start_time and end_time without migrating start_date and end_date? I had no idea that the database had BOTH and agree that one pair needs to be ditched. 

Looking at my copy of the database, it appears that start_time is  being used and start_date is completely empty. Therefore we should choose start_time over start_date, but I agree with @robkooper that we probably need a bit of time to verify that we're not using start_date or end_date anywhere in the code. Rob, this would be separate from PEcAn's internal use of variables named start_date and end_date since we use those to refer to datetime variables. Note that start_time is datetime and start_date is varchar so we're probably already OK.

db/migrate/20150209174734_add_drop_tables_and_columns.rb
gsrohde: Sure, go for it.  I was having writer's block when I named that.

gsrohde: How about modify_formats_runs_yields?

gsrohde:
@dlebauer I forget what our protocol is for marking a pull request as ready for me to review and pull.  Maybe "labelled with 'ready to review' AND assigned to me"?  And if so, how do I mark it when I'm not ready to accept a pull?  Assign it to some other person?

In any case, this pull request is still in the discussion stage while issues (not my issues, in this case) are being resolved, correct?

db/migrate/20150209174734_add_drop_tables_and_columns.rb
robkooper: sounds good.

gsrohde:
Rob: I renamed the migration (and did a few other tweaks).  Does this look OK now?

robkooper:
@mdietze this will work for BU library with a single update:
```
delete from mimetypes where id=1000000003 or id=1000000001;
```

robkooper:
+1

mdietze:
@robkooper Just deleted these two records on the BU server, but am a bit confused where they came from in the first place since the BETY web interface doesn't have a mimetypes tab and in the db version on our server the formats table is referencing mime_type by character not id

robkooper:
This migration will require a change to https://github.com/PecanProject/pecan/blob/master/db/R/dbfiles.R

robkooper:
@mdietze Not sure where they came from, maybe left over from testing the dbfiles function one day.

gsrohde:
Rob, is this ready to pull?  If so, assign it to me.

dlebauer:
@robkooper what changes will need to be made to dbfiles.R? Is there an issue for this in the PEcAn repository?

robkooper:
@dlebauer no there is no issue for this yet since I was unaware that columns are added/removed. Right now the code will add the mimetype to the dbfile, but will need to reference the values in mimetypes.

Easiest to see what happens is to install pecan (or the VM), do a git pull in bety and migrate the database. Next goto pecan and do a `scripts/build.sh --git --tests`. The tests should pass, and will hit a bunch of the database pieces.

gsrohde:
@robkooper I reassigned this to you since your last note says "I would like to wait with the pull request until we have the appropriate code for PEcAn."

dlebauer:
The appropriate changes for PEcAn are in pull request PecanProject/pecan#467

gsrohde:
@robkooper (and @dlebauer): 

What is the status of this?  Shall I pull it?

dlebauer:
from @robkooper 

> sure, needs to updated a bit to have all current code included.

Issue #203, opened 2015-01-30
Create API key for users


dlebauer:
* [x] generate api keys for existing users
* [x] show api key on users/show page
* [x] generate api key when new account is created
* [x] add key to email sent for new accounts

There is an apikey field in the users table, but only a few accounts have keys.

dlebauer:
can be closed after deploying #229

gsrohde:
Not really done—I haven’t merged pull request #229 let alone deployed it.  But I’ll work on merging it now.

dlebauer:
deployed in BETYdb 4.0

PR #229, opened 2015-02-13
create apikeys

phenolphtalein:
#203 
• added a button on the /users for admin to create api keys for all users who don't have one
• sends email to existing user after api key has been created
• added api key information on users/show page
• create api key when user signs up

### DIFF
--- a/app/views/contact_mailer/apikey_email.html.erb
+++ b/app/views/contact_mailer/apikey_email.html.erb
@@ -6,7 +6,7 @@
   <body>
     <h1>Hello, <%= @user[:name] %></h1>
     <p>
-      Your apikey for BETYdb is <%@user[:apikey]%>
+      Your apikey for BETYdb is <%= @user[:apikey]%>
     </p>
     <p>Thanks and have a great day!</p>
   </body>
--- a/app/views/contact_mailer/apikey_email.html.erb
+++ b/app/views/contact_mailer/apikey_email.html.erb
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Hello, <%= @user[:name] %></h1>
+    <p>
+      Your apikey for BETYdb is <%@user[:apikey]%>
+    </p>
+    <p>Thanks and have a great day!</p>
+  </body>
+</html>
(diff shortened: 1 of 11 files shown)


dlebauer:
* for new users, api key included in sign-up email

gsrohde:
@phenolphtalein , @dlebauer 

I guess this gets the job done, but it seems a little odd to me to create a button in the Web interface that you are probably only ever going to use once.

dlebauer:
@phenolphtalein 
* what is the purpose of the button? c
* an we generate api keys for existing users without sending an email?

phenolphtalein:
I will change the button to a script then. For the email, I thought we'd want to notify the existing users that an api key has been created for them.

dlebauer:
It would be nice to have a "generate api key" that will generate an api key and send to a user on a per-user basis.

That way, an individual user could change their apikey (e.g. if they accidentally compromised one by committing it to public code). And, we wouldn't need to generate an API key (and email) everyone on the list.

phenolphtalein:
Updated the page and removed the button to create all apikeys. Added a button to generate a new key for each individual user.

dlebauer:
Can you please document the rake task in the documentation? https://www.authorea.com/users/5574/articles/7062/_show_article

dlebauer:
close #203 when deployed

app/views/contact_mailer/apikey_email.html.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>Hello, <%= @user[:name] %></h1>
+    <p>
+      Your apikey for BETYdb is <%@user[:apikey]%>
```
gsrohde: This should be `<%= user[:apikey] %>`.

app/views/contact_mailer/apikey_email.txt.erb
```diff
@@ -0,0 +1,5 @@
+Hello, <%= @user[:name] %>
+
+Your apikey for BETYdb is <%@user[:apikey]%>
```
gsrohde: This should be `<%= user[:apikey] %>`.

gsrohde:
Please make fixes to the two email template files.

phenolphtalein:
@gsrohde  Fixed.

gsrohde:
@phenolphtalein  I have merged this, but I made some tweaks to DRY up the code and clean up the routes a bit.  (I also tweaked the e-mail templates a bit.)  If you're interested in seeing my changes, do
```
git diff 5de42f7 09b0b4d
```
and also look at the git log.

PR #281, opened 2015-04-22
add api_search support to trait and yields view

phenolphtalein:
Created an index action so traits_and_yields_view can be searched by the api (mainly used by the rOpensci package)


dlebauer:
@gsrohde please incorporate this into the next deployment (4.0 release?)

dlebauer:
works with `betydb.org/search.json?search=acer+rubrum` syntax

PR #301, opened 2015-05-27
Value constraints batch 1

gsrohde:
@dlebauer , @robkooper Please have a look at this and see if okay.

### DIFF
--- a/db/migrate/20150213162341_add_value_constraints_batch_1.rb
+++ b/db/migrate/20150213162341_add_value_constraints_batch_1.rb
@@ -298,7 +298,7 @@ ALTER TABLE dbfiles ADD CONSTRAINT valid_dbfile_container_type CHECK (container_
 ALTER TABLE ensembles ALTER COLUMN notes SET NOT NULL,
                       ALTER COLUMN notes SET DEFAULT '';
 ALTER TABLE ensembles ALTER COLUMN runtype SET NOT NULL,
-                      ADD CONSTRAINT valid_ensemble_runtype CHECK (runtype IN ('ensemble', 'sensitivity analysis', 'MCMC'));
+                      ADD CONSTRAINT valid_ensemble_runtype CHECK (runtype IN ('ensemble', 'sensitivity analysis', 'MCMC', 'pda.emulator'));
 
 /* ENTITIES */
 
@@ -458,9 +458,9 @@ ALTER TABLE runs ALTER COLUMN outprefix SET NOT NULL,
                  ALTER COLUMN outprefix SET DEFAULT '';
 ALTER TABLE runs ALTER COLUMN setting SET NOT NULL,
                  ALTER COLUMN setting SET DEFAULT '';
-ALTER TABLE runs ALTER COLUMN started_at SET NOT NULL;
-ALTER TABLE runs ADD CONSTRAINT valid_run_start_time CHECK (started_at <= NOW()::timestamp); -- NOW()::timestamp makes clear we are using local machine time
-ALTER TABLE runs ADD CONSTRAINT consistent_run_start_and_end_times CHECK (started_at <= finished_at AND finished_at <= NOW()::timestamp);
+-- ALTER TABLE runs ALTER COLUMN started_at SET NOT NULL;
+-- ALTER TABLE runs ADD CONSTRAINT valid_run_start_time CHECK (started_at <= NOW()::timestamp); -- NOW()::timestamp makes clear we are using local machine time
+-- ALTER TABLE runs ADD CONSTRAINT consistent_run_start_and_end_times CHECK (started_at <= finished_at AND finished_at <= NOW()::timestamp);
 COMMENT ON COLUMN runs.started_at IS 'system time when run was started';
 
 /* SITES */
@@ -593,7 +593,7 @@ ALTER TABLE users ALTER COLUMN country SET NOT NULL,
                   ALTER COLUMN country SET DEFAULT '';
 ALTER TABLE users ADD CONSTRAINT normalized_user_country_name CHECK (is_whitespace_normalized(country));
 ALTER TABLE users ALTER COLUMN crypted_password SET NOT NULL;
-ALTER TABLE users ADD CONSTRAINT valid_user_crypted_password_value CHECK (crypted_password ~ '^[0-9a-f]{40}$');
+ALTER TABLE users ADD CONSTRAINT valid_user_crypted_password_value CHECK (crypted_password ~ '^[0-9a-f]{1,40}$');
 /* ALTER TABLE users ALTER COLUMN salt SET NOT NULL; */
 ALTER TABLE users ALTER COLUMN access_level SET DATA TYPE level_of_access;
 ALTER TABLE users ALTER COLUMN page_access_level SET DATA TYPE level_of_access;
@@ -967,7 +967,8 @@ ALTER TABLE methods ALTER COLUMN description DROP NOT NULL;
 /* MODELS */
 
 ALTER TABLE models ALTER COLUMN model_name DROP NOT NULL;
-ALTER TABLE models ALTER COLUMN revision DROP NOT NULL;
+ALTER TABLE models ALTER COLUMN revision DROP NOT NULL,
+                   DROP CONSTRAINT normalized_revision_specifier;
 
 
 /* MODELTYPES_FORMATS */
@@ -1010,9 +1011,9 @@ ALTER TABLE runs ALTER COLUMN outprefix DROP NOT NULL,
                  ALTER COLUMN outprefix DROP DEFAULT;
 ALTER TABLE runs ALTER COLUMN setting DROP NOT NULL,
                  ALTER COLUMN setting DROP DEFAULT;
-ALTER TABLE runs ALTER COLUMN started_at DROP NOT NULL;
-ALTER TABLE runs DROP CONSTRAINT valid_run_start_time;
-ALTER TABLE runs DROP CONSTRAINT consistent_run_start_and_end_times; 
+-- ALTER TABLE runs ALTER COLUMN started_at DROP NOT NULL;
+-- ALTER TABLE runs DROP CONSTRAINT valid_run_start_time;
+-- ALTER TABLE runs DROP CONSTRAINT consistent_run_start_and_end_times; 
 COMMENT ON COLUMN runs.started_at IS 'system time when run begins';
 
 /* SITES */


gsrohde:
The 3 migrations are probably the main things to look at.

robkooper:
Vacation this week. Will check next week, need to make sure I can update BU database first to current version of BETY.

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1194 @@
...
+ALTER TABLE dbfiles ADD CONSTRAINT valid_dbfile_container_type CHECK (container_type IN ('Model','Posterior','Input'));
+-- consistency checks between this table and inputs, models, and posteriors tables are part of a separate migration
+
+
+/* ENSEMBLES */
+
+ALTER TABLE ensembles ALTER COLUMN notes SET NOT NULL,
+                      ALTER COLUMN notes SET DEFAULT '';
+ALTER TABLE ensembles ALTER COLUMN runtype SET NOT NULL,
+                      ADD CONSTRAINT valid_ensemble_runtype CHECK (runtype IN ('ensemble', 'sensitivity analysis'));
```
robkooper: runtype can now also be MCMC

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1194 @@
...
+
+
+
+/* This is based on http://www.w3.org/Addressing/URL/5_BNF.html */
+CREATE OR REPLACE FUNCTION is_host_address(
+  string text
+) RETURNS boolean AS $body$
+DECLARE
+  xalphas text := $$([-\w$@.&+!*"'(),]|%[0-9a-f]{2})+$$;
+  ialpha text := '[a-z]' || xalphas;
```
robkooper: need to allow for upper case letters, i.e. [a-zA-Z]

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1194 @@
...
+/* MIMETYPES */
+
+-- Constraints are in AddUniquenessConstraints migration.
+
+
+/* MODELS */
+
+ALTER TABLE models ALTER COLUMN model_name SET NOT NULL,
+                   ADD CONSTRAINT no_spaces_in_model_name CHECK (is_whitespace_free(model_name));
+ALTER TABLE models ALTER COLUMN revision SET NOT NULL,
```
robkooper: why no spaces in revision?

gsrohde: Currently these are all numbers, the empty string or 'unk'.  But we can relax this restriction if you like ( whitespace-normalized?).  Say the word.

robkooper: Some of the models that Mike has in his database have spaces, so yes please.

db/migrate/20150213162341_add_value_constraints_batch_1.rb
gsrohde: @robkooper All caps, just like that?

robkooper: That is how it is right now in the BU database.

gsrohde:
@robkooper I updated this pull request, addressing the three concerns you had.  In the process, I noticed a few further issues.  These all have to do with what your update/migrate scripts do with the users table.

1. The user1, user2, etc. users get a `crypted_password` value of length 32 rather than 40.  The also get a `salt` value of 'nosecret' instead of a 40 digit hexidecimal number.

2. The "caryaXX" and "guestuser" users get salt values of 'carya' and 'bety' respectively instead of 40 digit hexidecimal numbers.

3. The `apikey`, `stat_prov`, and `postal_code` columns are set to NULL rather than the empty string or some bona fide value.

If you don't want to bother changing your scripts, let me know, and I'll relax the restrictions on the users table columns.

robkooper:
guestuser is set by BETY, the crypted password was generated using the salt. If this is changed we will need to update PEcAn to reflect this. Are you guaranteed to always get a hash of 40 chars for the password? If possible I would leave it as is.

robkooper:
I think NULL is a better value to indicate that it is not set (instead of set to an empty string).

gsrohde:
The encryption algorithm always returns the result of a call to Digest::SHA1.hexdigest in the standard Ruby library ‘digest’ which I’m nearly certain always produces an 40 char return value no matter what arguments it is given.  Presumably, then, it is impossible to log in in as a user if the crypted_password value is anything other than 40 hex characters.  (As an aside, the algorithm seems to always produce values of Ruby type “String" and seems to always use lower case hex digits.  In other words, log-in will fail if crypted_password is replaced by UPPER(crypted_password).  This is a reason for using '[0-9a-f]’ rather than '[[:xdigit:]]’ in the check constraint.  This to me seems like a sort of defect in the authentication library.  Or perhaps it only means that we chose the wrong data type for crypted_password in the users table.)

The upshot of this is that constraining crypted_password to 40 lower case hex digits seems reasonable to me since any other value will be non-functional.  But it doesn’t seem terribly important that we do this, so I won’t be a stickler about it, especially if involves a lot of extra work for you.

Although the authentication library also always generates salt values that are 40 character lower case hex strings, any value for salt will work (by which I mean a functional password—crypted_password pair is possible with the given salt value); and so it seems reasonable to drop the restriction I wanted to impose and allow such values as ‘carya’ and ‘nosecret’ since they “work”.

gsrohde:
"I think NULL is a better value to indicate that it is not set (instead of set to an empty string)."

Are you speaking of `apikey` in particular, or do you mean `stat_prov` and `postal_code` as well?

In general I don't agree, but I'm more concerned about the use of NULLs in tables the represent real data or that may be involved in complex queries, so again, if it is a pain to change your scripts to produce strings, I don't have a problem with dropping the NOT NULL restriction on these columns.  Let me know.

robkooper:
I can do either pick one. Just make sure that if the apikey is set to "" the user can not login with an empty apikey (which is why i think NULL is better in this case meaning no API key vs an empty API key).

gsrohde:
Good point with apikey--I think '' would open up access to anyone who appends `?key=` to a URL unless we took steps to prevent that.  So I'll enforce length = 40 hex or NULL.

I'll keep NOT NULL on `stat_prov` and `postal_code` and the current restriction on `crypted_password` and relax the restriction on salt ... unless you veto any of these decisions as creating too much of a headache for you or for uses of the sync script.

gsrohde:
@robkooper I'm going to plan on pulling this and maybe even deploying this tomorrow under the assumption that you've voiced all your concerns and that I've addressed them to your liking.

Please let me know ASAP if you have any further concerns or if you need more time to change your scripts should it be necessary.  (Or if that's going to take a deal of time, I can perhaps drop the problem restrictions, at least for now.)

gsrohde:
@robkooper I just thought I should also alert you to the fact that whatever process is inserting into the `runs` table is inserting rows that violate these new constraints.  Except for the `started_at` column, which won't have a default value, the other columns may not be a problem: together with the new NOT NULL constraint for `outprefix` and `setting`, there is a new default value of '' for these as well.  It depends of course on whether the insert statements leave out these columns or explicitly set them to NULL.

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1311 @@
...
+
+/* RUNS */
+
+ALTER TABLE runs ALTER COLUMN outdir SET NOT NULL,
+                 ALTER COLUMN outdir SET DEFAULT '';
+ALTER TABLE runs ALTER COLUMN outprefix SET NOT NULL,
+                 ALTER COLUMN outprefix SET DEFAULT '';
+ALTER TABLE runs ALTER COLUMN setting SET NOT NULL,
+                 ALTER COLUMN setting SET DEFAULT '';
+ALTER TABLE runs ALTER COLUMN started_at SET NOT NULL;
```
robkooper: the started_at and finished_at can be null, when the run is created these are not set. These are set when the run is actually started and finished.

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1311 @@
...
+ALTER TABLE users ADD CONSTRAINT well_formed_user_email CHECK (is_wellformed_email(email));
+ALTER TABLE users ALTER COLUMN city SET NOT NULL,
+                  ALTER COLUMN city SET DEFAULT '';
+/* FIX */ --UPDATE users SET city = normalize_whitespace(city) WHERE NOT is_whitespace_normalized(city);
+ALTER TABLE users ADD CONSTRAINT normalized_user_city_name CHECK (is_whitespace_normalized(city));
+ALTER TABLE users ALTER COLUMN country SET NOT NULL,
+                  ALTER COLUMN country SET DEFAULT '';
+ALTER TABLE users ADD CONSTRAINT normalized_user_country_name CHECK (is_whitespace_normalized(country));
+ALTER TABLE users ALTER COLUMN crypted_password SET NOT NULL;
+ALTER TABLE users ADD CONSTRAINT valid_user_crypted_password_value CHECK (crypted_password ~ '^[0-9a-f]{40}$');
```
robkooper: I thought the plan was to drop the 40 chars? the php code I use does not have 40 chars password generated. I think it might depend on the salt.

db/migrate/20150213162341_add_value_constraints_batch_1.rb
```diff
@@ -0,0 +1,1311 @@
...
+ALTER TABLE dbfiles ADD CONSTRAINT valid_dbfile_container_type CHECK (container_type IN ('Model','Posterior','Input'));
+-- consistency checks between this table and inputs, models, and posteriors tables are part of a separate migration
+
+
+/* ENSEMBLES */
+
+ALTER TABLE ensembles ALTER COLUMN notes SET NOT NULL,
+                      ALTER COLUMN notes SET DEFAULT '';
+ALTER TABLE ensembles ALTER COLUMN runtype SET NOT NULL,
+                      ADD CONSTRAINT valid_ensemble_runtype CHECK (runtype IN ('ensemble', 'sensitivity analysis', 'MCMC'));
```
robkooper: Can you add 'pda.emulator' as a new type, this just got introduced.

PR #328, opened 2015-06-25
Fix return type of id-getting functions to return bigint rather than …

gsrohde:
…int.


gsrohde:
@robkooper 

Does this look okay?

robkooper:
@gsrohde looks good and works.

PR #330, opened 2015-06-26
Adding script for creating an SQL script for inserting new management…

gsrohde:
…s from a

CSV file describing the managements to be inserted.


gsrohde:
@dlebauer 

You can try this out by checking out this branch.

You will have to run bundle install first since there is one new Gem.

Run from the Rails root directory with
```
script/insert_managements.rb --help
```

Once you specify the proper arguments, this will work with your local copy of BETYdb.  I'll tell you later how to set it up to work with ebi_production on ebi-forecast.

See if the help message is clear enough to get you started, and give me feedback if it isn't clear.

gsrohde:
@dlebauer 

To run the script when gems are installed in vendor/bundle instead of the system location, use `bundle exec`:
```
bundle exec script/insert_managements.rb --help
```
Do this and you won't need the `require_relative` hack.

Even if you _aren't_ using gems out of vendor/bundle, you should use `bundle exec` anyway for any Ruby script (such as this one) that loads the Rails environment.  This will ensure the script sees the same environment (in particular, the same gemset) that the Rails Web application sees.  (However, `bundle exec` is not necessary if you are using RVM.)

PR #351, opened 2015-09-24
fix wrong link

### LINKED ISSUE #332: Issue with reloading workflows through web BETYdb
Through the PEcAn web interface a workflow result/history page is located at: http://test.esh.bnl.gov/pecan/08-finished.php?workflowid=2000000002

However, if I log on to the BETY interface on our server the link on the http://test.esh.bnl.gov/bety/workflows page takes me to: http://test.esh.bnl.gov/pecan/finished.php?workflowid=2000000002

This results in a failed recall of the run history.  Note the missing "08-" prefix.  Is this a bug?  Or do I have a config problem?

@robkooper @dlebauer @gsrohde

robkooper:
Fixes #332

### DIFF
--- a/app/views/workflows/_index_table.html.erb
+++ b/app/views/workflows/_index_table.html.erb
@@ -24,7 +24,7 @@
     <td class="center"><%=h workflow.created_at %></td>
     <td class="center"><%=h workflow.updated_at %></td>
 
-    <td class="center"><%= link_to 'Results', "/pecan/finished.php?workflowid=" + workflow.id.to_s %></td>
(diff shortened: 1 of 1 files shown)


gsrohde:
@robkooper I don't think this solution is going to work properly at www.betydb.org.  If you want, I'll take this over and try to get this to work on the live site.  Also, if you want things to work with other deployments, I can go ahead and merge this--it won't make things any worse.

robkooper:
If possible get this in, and we can try to figure out a better way to do it. Most sites will have it setup with /bety and /pecan, but I agree we should have a better way to configure what pecan to link to.

PR #358, opened 2015-10-12
add notes and user to workflow

robkooper:
update web to show notes and user


robkooper:
See https://github.com/PecanProject/pecan/issues/134

PR #359, opened 2015-10-12
add log option to rakefile

robkooper:
Now can run rake log db:migrate to see exact sql statement executed.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,4 +4,8 @@
 require File.expand_path('../config/application', __FILE__)
 require 'rake'
 
+task :log => :environment do
+  ActiveRecord::Base.logger = Logger.new(STDOUT)
+end
+
 BetyRails3::Application.load_tasks
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,9 @@
 require File.expand_path('../config/application', __FILE__)
 require 'rake'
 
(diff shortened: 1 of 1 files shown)


robkooper:
This helps to make sure the right SQL statements are executed, and allow the SQL statements to be shared with somebody that only cares about updating the database when working on PEcAn or somebody that would like to run the migrations manually.

When running the following `rake log db:migrate` it will spit out:

```
Migrating to AddNotesUserToWorkflow (20151011190026)
   (0.0ms)  BEGIN
==  AddNotesUserToWorkflow: migrating =========================================
-- add_column(:workflows, :notes, :text)
   (0.4ms)  ALTER TABLE "workflows" ADD COLUMN "notes" text
   -> 0.0006s
-- add_column(:workflows, :user_id, :integer, {:limit=>8})
   (0.1ms)  ALTER TABLE "workflows" ADD COLUMN "user_id" bigint
   -> 0.0002s
==  AddNotesUserToWorkflow: migrated (0.0009s) ================================

   (0.2ms)  INSERT INTO "schema_migrations" ("version") VALUES ('20151011190026')
   (1.8ms)  COMMIT
   (0.5ms)  SELECT version FROM schema_migrations ORDER BY version
```

PR #369, opened 2015-11-18
Added py script for creating PostGIS queries

max-zilla:
To streamline process of inserting a set of coordinates as a Polygon to
sites.geometry in BETY

### DIFF
--- a/script/buildSQLGeom.py
+++ b/script/buildSQLGeom.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python
+
+"""
+This module will accept an input text file with (x,y,z) coordinates in a CSV-format list
+and generate a SQL string that will add these coordinates as a PostGIS Polygon
+to sites.geometry within a BETYdb instance.
+
+Command line usage:
+        -h      show help
+        -i      input filename
+        -e      EPSG reference number, default is 3857
+        -s      site_id in BETYdb sites table to associate geometry with
+        -o      flag to write resulting SQL to output file, as "<input_filename>_SQL.txt"
+        -x      will attempt to execute resulting SQL on db instance defined in next 4 params
+        -n      PostgreSQL host address
+        -d      Database name, default is 'bety'
+        -u      Username
+        -p      Password
+
+Example:
+        buildSQLGeom.py -i C:/folder/coordinates.csv -e 4326 -s 19000000000 -x -n localhost -d bety -u GUEST -p GUES -o
+
+Sample input file contents:
+        LONGITUDE,LATITUDE,ALTITUDE
+        -76.116081,42.794448,415
+        -76.116679,42.794448,415
+        -76.116679,42.79231,415
+A header row should be included in the document, although the column names do not matter.
+"""
+
+import sys
+import getopt
+#import psycopg2
+
+def executeSQL(query, host, dbname, user, passwd):
+        """
+        Execute SQL query on specified host database with given credentials.
+                Requires psycopg2: https://pypi.python.org/pypi/psycopg2 (un-comment line 32!)
+                Adapted from https://wiki.postgresql.org/wiki/Using_psycopg2_with_PostgreSQL
+        :param query: SQL string containg query to execute.
+        :param host: Host address of target PostgreSQL instance
+        :param dbname: Database to execute query in
+        :param user: Username
+        :param passwd: Password
+        """
+
+        # Build connection string & remove excess quotes if given
+        conn_string = "host='"+host+"' dbname='"+dbname+"' user='"+user+"' password='"+passwd+"'"
+        conn_string = conn_string.replace("=''", "='").replace("'' ", "' ")
+        print "Connecting to database\n ->%s" % (conn_string)
+
+        try:
+                conn = psycopg2.connect(conn_string)
+                cursor = conn.cursor()
+                print("Sucessfully connected.")
+                cursor.execute(query)
+                
+        except Exception as e:
+                print(e)
+
+def main(argv):
+        # These are default values, will be overwritten by command line parameters if given.
+        input_file = None
+        epsg       = 3857
+        site_id    = 0
+        out_file   = False
+        run_query = False
+        host = None
+        dbname = 'bety'
+        user = 'USERNAME'
+        passwd = 'PASSWORD'
+
+        # Parse command line parameters
+        try:
+                opts, args = getopt.getopt(argv, "hi:e:s:xn:d:u:p:o", ["input=","output=","epsg=","site=","execute","host=","dbname=", "user=", "pass="])
+        except getopt.GetoptError:
+                print('buildSQLGeom.py -i <inputfile> -e <epsg> -s <site_id> -x -n <postgres_hostname> -d <dbname> -u <username> -p <password> -o <outputfile> ')
+                print('epsg default is 3857')
+                print('-o will write query to <input_filename>_SQL.txt')
+                print('-x will attempt to execute resulting query')
+                print('-n, -d, -u, -p required if -x flag enabled')
+                sys.exit(2)
+        for opt, arg in opts:
+                if opt == '-h':
+                        print('buildSQLGeom.py -i <inputfile> -e <epsg> -s <site_id> -x -n <postgres_hostname> -d <dbname> -u <username> -p <password> -o <outputfile> ')
+                        print('epsg default is 3857')
+                        print('-o will write query to <input_filename>_SQL.txt')
+                        print('-x will attempt to execute resulting query')
+                        print('-n, -d, -u, -p required if -x flag enabled')
+                        sys.exit()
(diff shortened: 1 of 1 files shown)


script/buildSQLGeom.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
+
+"""
+This module will accept an input text file with (x,y,z) coordinates in a CSV-format list
```
dlebauer: Can we either make z optional or query elevation from a public API?

It appears this can be done with gdal http://gis.stackexchange.com/a/29639/3218; @yanliu-chn do have any suggestions?

script/buildSQLGeom.py
```diff
@@ -0,0 +1,153 @@
...
+and generate a SQL string that will add these coordinates as a PostGIS Polygon
+to sites.geometry within a BETYdb instance.
+
+Command line usage:
+        -h      show help
+        -i      input filename
+        -e      EPSG reference number, default is 3857
```
dlebauer: shouldn't default be 4356 / WGS84 (#96). I don't know if this is a good idea, but I believe it is the default in BETYdb.

script/buildSQLGeom.py
```diff
@@ -0,0 +1,153 @@
...
+        -s      site_id in BETYdb sites table to associate geometry with
```
dlebauer: could the `-s` argument accept either integer or string, and if a string, either query the site_id (if exists) or add a new site (if it doesn't exist)?

script/buildSQLGeom.py
```diff
@@ -0,0 +1,153 @@
...
+        user = 'USERNAME'
+        passwd = 'PASSWORD'
+
+        # Parse command line parameters
+        try:
+                opts, args = getopt.getopt(argv, "hi:e:s:xn:d:u:p:o", ["input=","output=","epsg=","site=","execute","host=","dbname=", "user=", "pass="])
+        except getopt.GetoptError:
+                print('buildSQLGeom.py -i <inputfile> -e <epsg> -s <site_id> -x -n <postgres_hostname> -d <dbname> -u <username> -p <password> -o <outputfile> ')
+                print('epsg default is 3857')
+                print('-o will write query to <input_filename>_SQL.txt')
```
dlebauer: rename to something like `<input_filename>_insert.sql`

script/buildSQLGeom.py
```diff
@@ -0,0 +1,153 @@
...
+        l = csv.readline().rstrip() # first data line
+        while l:
+                # Get values from each row of file separated by ',' and append to query
+                coords = l.split(",")
+                q_line = coords[0] + " " + coords[1] + " " + coords[2] + ","
+                query += q_line 
+                l = csv.readline().rstrip()
+
+        # Finish the end of the query and print to console
+        query += "))', "+str(epsg)+") WHERE ID="+str(site_id)
```
dlebauer: `ST_GeomFromText` seems to require srid http://postgis.net/docs/ST_GeomFromText.html

max-zilla: For clarity I've changed var names from EPSG to be SRID, and made the default 4326. The command line argument is now -r, for (spatial) reference ID and the docs are updated accordingly.

script/buildSQLGeom.py
max-zilla: This is no problem - we'll just need to make sure the user understands that if a sitename is provided, a valid BETY host must also be defined so that the site_id can be queried or created.

script/buildSQLGeom.py
max-zilla: We can certainly make it optional. For querying, the example you point assumes we have a raster of elevation that we're querying.

However, we may be in luck that the USGS National Elevation Dataset has an API:
http://ned.usgs.gov/epqs/

Here is an example query that returns elevation for XY (-76.116081,42.794448):
http://ned.usgs.gov/epqs/pqs.php?x=-76.116081&y=42.794448&units=Meters&output=xml
Elevation can be feet or meters.

dlebauer: **meters** Always use SI by default. NED at 1/3 arc second is ~10m thats great. Outside the domain, throw a warning and that querying elevation outside the US is not yet implemented, with a link to 'request this feature at https://github.com/pecanproject/bety/issues/new' unless there is an even easier alterinative (@yanliu-chn?)

max-zilla: I've got a working path to the USGS API implemented, but now I've come across the google elevation API which has global support. 
https://developers.google.com/maps/documentation/elevation/intro

We'd need a [google maps API key](https://developers.google.com/maps/documentation/elevation/get-api-key) and are subject to traffic quota constraints, but this supports passing a set of points (e.g. a path or line) and returning sampled elevations from along that path. It'll even generate altitude histograms for you if we wanted to get cute.

My guess is that since this is for open research the API key should be no problem and since we're not building some kind of high-frequency mobile app, I doubt traffic volume would be a concern.

yanliu-chn: Please try usgs query service http://nationalmap.gov/epqs/ for US locations. It's quite easy to figure out a http get request to this service using code.

script/buildSQLGeom.py
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env python
+
+"""
+This module will accept an input text file with (x,y,[z]) coordinates in a CSV-format list
+and generate a SQL string that will add these coordinates as a PostGIS Polygon
+to sites.geometry within a BETYdb instance.
+
+A header row should be included in the input file, although the column names do not matter. If a z column
+for altitude is not included, the script will attempt to get it from USGS NED (US) or Google (global).
```
dlebauer: is the google elevation api implemented as indicated by these comments? it is not necessary, but I don't see it in the script

max-zilla: No, not yet. I need to get an API key registered first, will be smooth sailing then.

max-zilla: I just committed a version with the google elevation API. However, I am not using it by default - it requires the user to point to a text file with a valid Google Maps API key (they use for traffic quotas). 

I can provide you with my API key if you like, or you can generate your own:
https://developers.google.com/maps/documentation/elevation/get-api-key

If that file is present and **line 130** points to it, the API should work and we can call it instead of the USGS API on **line 233**.

dlebauer: I have a key.. But to confirm, points in US query USGS by default, correct?

max-zilla: In the current code, *everything* will go through USGS. My thinking was that if you have a valid key, we could just make everything go through Google. Do you consider USGS the preferred authority in the US?

We can make USGS domestic default and Google international default, but we need a good way to determine whether the point is in the US to begin with, which is likely another API call. Perhaps the anticipated traffic is low enough that this is fine. 

Another option would be to query USGS, and if that doesn't find anything, we query Google. Or, we could define a simple rectangle around the US and quickly check whether the lat/lon exceed that boundary, but that falls apart e.g. in northern Mexico. Don't love that idea.

dlebauer: USGS is preferred because it is more accurate and consistent. The accuracy
of Google results is inversely proportional to the number of points.

Id go with 1. Query USGS. 2. If no data query Google.

dlebauer:
@max-zilla I just tried this out and got 

```
[script]$ python buildSQLGeom.py -h
  File "buildSQLGeom.py", line 77
    except Exception as e:
                      ^
SyntaxError: invalid syntax
```

script/buildSQLGeom.py
```diff
@@ -0,0 +1,287 @@
...
+        -o      flag to write resulting SQL to output file, as "<input_filename>_insert.sql"
+        -e      will attempt to execute resulting SQL on db instance defined in next 4 params
+        -n      PostgreSQL host address
+        -d      Database name, default is 'bety'
+        -u      Username
+        -p      Password
+
+Examples:
+        buildSQLGeom.py -i C:/folder/coordinates.csv -r 4326 -s 19000000000 -e -n localhost -d bety -u GUEST -p GUES -o
+        buildSQLGeom.py -x 76.116081 -y 42.794448
```
dlebauer: @max-zilla in this example, no site_id or sitename is provided. Does this create a new site?

max-zilla: Currently, if not site_id or sitename is provided, site_id is set to 0. I can modify it to create a new site, but from what I understand of [the sites table](https://www.betydb.org/db_docs/tables/sites.html) we would need these non-nullable fields:
* city
* state
* country
* soil
* notes
* soilnotes
* sitename

I'll augment the examples to include one of those arguments.

max-zilla: As for the error, I forgot to remove that Exception placeholder as I've not yet tested the SQL execution on a live BETY instance (should be able to do that this week). I'll modify code in the meantime to catch all exceptions so you can at least see what error you're encountering there.

PR #371, opened 2015-11-30
Changed SQL Exception behavior

max-zilla:


gsrohde:
@max-zilla I made a pull-request to your master branch (https://github.com/max-zilla/bety/pull/1) having some suggested changes and lots of comments.  I don't necessarily mean for you to pull in my changes wholesale, but I thought it was a convenient way to address concerns and offer suggestions.  Also, I'm going to add some comments below on lines that aren't in the diff.

gsrohde:
Line 8: "column names do not matter": It would be more robust to make the names matter but the order not matter.  See comment below about using the CSV module.

Line 9: I couldn't get altitude lookup to work.

Line 135: Should this file be checked in as well?

Line 226: You should only construct the argument to `ST_Geomfromtext()` here instead of the whole query.  (See note on line 351 of my pull request.)



Line 230: Although I've never used it, I think the Python Standard Library module `csv` would simplify much of the code here (and make it more robust).

dlebauer:
If I try this

```sh
curl -O https://raw.githubusercontent.com/max-zilla/bety/master/script/buildSQLGeom.py
printf "X,Y\n38.0,-97.5\n38.0, -95.5\n40.0, -95.5\n40.0, -97.5\n38.0, -97.5.0" > manhattan_ks.csv
python26 buildSQLGeom.py -i manhattan_ks.csv -s "Manhattan, KS region" -d "bety" -u "bety" -p "bety" -o manhattan_ks.sql
```

I get the following error:

> site_id cannot be queried without BETYdb credentials. using 0 as default.
found longitude in column 0: "X"
found latitude in column 1: "Y"
Traceback (most recent call last):
  File "buildSQLGeom.py", line 345, in <module>
    main(sys.argv[1:])
  File "buildSQLGeom.py", line 277, in main
    alt = getUSGSAltitude(lon, lat)
  File "buildSQLGeom.py", line 119, in getUSGSAltitude
    return alt_req.json()['USGS_Elevation_Point_Query_Service']['Elevation_Query']['Elevation']
  File "/usr/lib/python2.6/site-packages/requests/models.py", line 604, in json
    return json.loads(self.text or self.content)
  File "/usr/lib64/python2.6/json/__init__.py", line 307, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python2.6/json/decoder.py", line 319, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python2.6/json/decoder.py", line 338, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded

even though the connection parameters appear to be valid:

```sh
psql  --dbname=bety --username=bety --password
```

lets me in

dlebauer:
It works now using `-n localhost` and providing an elevation `-z 300`

gsrohde:
@max-zilla When you want me to look at this—when you think it is ready to be pulled—assign it to me.

max-zilla:
Will do @gsrohde. I pushed some fixes but I'd like to see if @dlebauer still encounters any issues on this file he's looking at before declaring it ready.

dlebauer:
@max-zilla when I do this

```sh
printf "X,Y\n38.0,-97.5\n38.0, -95.5\n40.0, -95.5\n40.0, -97.5\n38.0, -97.5.0" > manhattan_ks.csv
python buildSQLGeom.py -i manhattan_ks.csv -s "Manhattan, KS region" -d terra-test -u bety -p bety -n localhost
```

1. It does look up the correct site but doesn't do the insert
2. the update statement it gives does not include altitude

> UPDATE sites SET geometry = ST_Geomfromtext('POLYGON((38.0 -97.5,38.0 -95.5,40.0 -95.5,40.0 -97.5,38.0 -97.5.0,38.0 -97.5))', 4326) WHERE ID=7000000016

dlebauer:
P.S. its great to be able to bypass USGS if an error or unavailable, but since both of these sites are in the US, it might be worthwhile to figure out why the USGS elevation data is not queried.

max-zilla:
@dlebauer a couple things about your manhattan_ks file.

- X and Y are swapped - that's why USGS wasn't getting altitude. You have lat in the X column and vice versa. Swapping these was returning an altitude from USGS properly.

- your final coordinate has a typo - "-97.5.0". That extra ".0" is the parse error you're seeing I think.

When I adjust those two I get:
```
UPDATE sites SET geometry = ST_Geomfromtext('POLYGON((-97.5 38.0 422.885559,-95.5 38.0 422.885559,-95.5 40.0 422.885559,-97.5 40.0 422.885559,-97.5 38.0 422.885559))', 4326) WHERE ID=7000000016
```

Issue #374, opened 2015-12-09
Editing a Site with a non-Point geometry will result in data-loss


gsrohde:
The geometry will silently be converted from a region to a point, even if you don't actually change anything before hitting "Update."

Need advice on what the behavior _should_ be.  Possibly: Don't allow updating non-Point geometries via the Rails app, and leave them alone when editing other attributes of a Site.

dlebauer:
For now, it will be prudent to disable the interface so that this bug does not occur, i.e.

> Don't allow updating non-Point geometries via the Rails app

But a better solution would be to allow a user to copy/paste or upload either the two or three column dataset (lat, lon, elevation) that can be read by the script that @max-zilla wrote (#369, #373). (would this be an appropriate use of a clowder extractor?).

gsrohde:
@dlebauer See proposed fix at http://pecandev.igb.illinois.edu/sr/.

dlebauer:
@gsrohde comments in pull request #375

gsrohde:
@dlebauer Release 4.5 fixes this bug but doesn't provide a way to enter non-point geometries.  I recommend closing this and if you want, make a new issue for a feature request.

Issue #373, opened 2015-12-08
"Elevation" in sites show page should not be NaN


dlebauer:
I added a polygon to the sites in ebi-forecast.igb.illinois.edu/bety-mepp using

```sql
 UPDATE sites SET geometry = ST_Geomfromtext('POLYGON((40.062806716 -88.204463395 
  640.0,40.062776529 -88.206952545 640.0,40.063609417 -88.206951006 640.0,40.063617384  
  -88.205248163 640.0,40.063849512 -88.204853694 640.0,40.063896590 -88.204479037 
  640.0,40.062806716 -88.204463395 640.0))', 4326) WHERE ID=8000000002;
```

However, when I look at the show page for this record, the 'elevation' field is given as `NaN`. This is confusing since I specified the elevation.

gsrohde:
@dlebauer The problem is that when dealing with non-Point geometries, the Rails sites pages (as we currently have them set up) first take the centroid of the geometry before extracting coordinates.  And ST_Centroid returns a two-dimensional point.

And this behavior probably makes sense: Consider the case where the geometry describes the perimeter of a hill.  If you based the displayed altitude on the altitude of the perimeter points, it would be much lower than the altitude of the center of the area.

Even in the case where the area is a flat but sloping surface so that the altitude of the center is in fact somewhere in between the altitudes of the perimeter points, it is not clear how the altitudes should be averaged.  Suppose we give them equal weight.  And suppose the lower perimeter of the area is much more ragged than the upper perimeter so that many more points are needed to describe it.  Then the average altitude we compute is likely to be significantly lower than the altitude of the center.

If the common use case for which you want to show the altitude of a region is for a flat field, then we could make a special case whereby _if_ the perimeter points all have the same altitude, then that will be taken as the representative altitude of the region; and otherwise the display for the altitude of a non-Point geometry could say "indeterminate" or something.  (By the way, I discussed this all with Max, and he suggested he could look in to the existence of libraries that might compute some sort of 3-D centroid, if it comes to that.  But again, doing so seems to me highly problematic for the afore-mentioned reasons.)

**This all brings up what is probably a much more serious issue:** If you go into the Edit Site page for some site that has a non-Point geometry, and if you hit "Edit" (even without editing anything) **the Rails app will silently replace your region with a point!** (a 3-D point corresponding to the centroid but with altitude set to zero, I think).  I'm about to make a separate issue for this.

[cc @max-zilla]

dlebauer:
[...]

PR #375, opened 2015-12-11
Fixes for Site-geometry-releated problems

gsrohde:
Fixes for GH issues #373 and #374.

### DIFF
--- a/app/helpers/sites_helper.rb
+++ b/app/helpers/sites_helper.rb
@@ -9,6 +9,8 @@ module SitesHelper
     if @site.point?
       options = common_options
     else
+      # Use :dummy (which returns "") as the attribute to force display of the placeholder.
+      attribute = :dummy
       options = common_options.merge(disabled: true,
                                      placeholder: "Site coordinates can't be updated for non-point sites.")
     end
--- a/app/helpers/sites_helper.rb
+++ b/app/helpers/sites_helper.rb
@@ -9,10 +9,7 @@ module SitesHelper
     if @site.point?
       options = common_options
     else
-      # Use :dummy (which returns "") as the attribute to force display of the placeholder.
-      attribute = :dummy
-      options = common_options.merge(disabled: true,
-                                     placeholder: "Site coordinates can't be updated for non-point sites.")
+      options = common_options.merge(disabled: true)
     end
 
     site_form.text_field(attribute, options)
--- a/app/helpers/sites_helper.rb
+++ b/app/helpers/sites_helper.rb
@@ -13,7 +13,7 @@ module SitesHelper
                                      placeholder: "Site coordinates can't be updated for non-point sites.")
     end
 
-    text_field(site_form, attribute, options)
+    site_form.text_field(attribute, options)
 
   end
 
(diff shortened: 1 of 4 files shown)


gsrohde:
@max-zilla Look over this code if you want to start getting into some Rails code.

@dlebauer I have a working copy of this branch at http://pecandev.igb.illinois.edu/sr/  See if this looks OK to you.  An example site having non-point geometry is 1202.

app/helpers/sites_helper.rb
```diff
@@ -1,5 +1,24 @@
...
+  # point.
+  def geometry_aware_text_field(site_form, attribute, common_options)
+
+    if @site.point?
+      options = common_options
+    else
+      # Use :dummy (which returns "") as the attribute to force display of the placeholder.
+      attribute = :dummy
+      options = common_options.merge(disabled: true,
+                                     placeholder: "Site coordinates can't be updated for non-point sites.")
```
dlebauer: Only suggestion would be to put the centroid values in, and have a note at the top that says something more like "non-point site; use SQL to update or contact administrator for more information"

gsrohde:
@dlebauer See if you like what I now have up at  http://pecandev.igb.illinois.edu/sr/.

dlebauer:
@gsrohde looks good. Except that the warning also shows up for point sites, e.g. 

http://pecandev.igb.illinois.edu/sr/sites/76/edit

Thanks!

gsrohde:
I think this was a case of the warning message persisting to other pages after you go to a non-point site.  Check if it seems fixed now.

dlebauer:
fixed now. Thanks!

dlebauer:
@max-zilla I'll assign to you to review and merge.

gsrohde:
@max-zilla Wait a sec until I merge my latest fixes into this branch.

gsrohde:
@max-zilla OK, this is ready to review and merge.

app/models/site.rb
```diff
@@ -62,12 +62,17 @@ def lon=(val)
     end
   end
 
+  # Returns true if the site geometry is null or has type 'ST_Point'.
+  def point?
+    geometry.nil? || geometry.geometry_type.type_name == 'Point'
+  end
+
```
max-zilla: If this is meant to return true if geometry is a point, why is it true if geometry is nil? Is this primarily a way to ensure the geometry is not a line or complex shape?

max-zilla:
@gsrohde I have a little question about point? definition but otherwise I think the rest makes sense.

@dlebauer I don't have write access to the repo so I won't be able to merge this pull request, but I have reviewed the changes.

app/models/site.rb
gsrohde: It is primarily a way to ensure we don't destroy geographical data from non-point sites by replacing them with their centroid when they are edited via Rails, which is what happens now.  We don't have to worry about destroying data if the geometry is already a point or if it is null.  I guess you could argue that "point?" is misnamed since if it's NULL it's not really a point, but this is no more offensive it seems than SQL's generally cavalier semantics surrounding the use of NULL.

The Rails code currently prohibits saving a site that doesn't have geographical data, but this hasn't always been the case (otherwise this kludge would be unnecessary—this is one reason I've been pushing hard for sensible database constraints over the past year or so).  26 out of the 1049 sites in the ebi_production database have NULL in the geometry column.

max-zilla: Makes perfect sense. You addressed what was essentially the subtext of my question - "is the name misleading?"

Issue #377, opened 2015-12-11
Allow new users to request changes to their own access levels


dlebauer:
When I sign up with default access level, then edit my user record, I can select page access Administrator and data access Restricted.

The interface says "Successfully updated" even though records are not updated. The user/edit interface should be the same as the user/new interface in that it should send a request to the administrator(s).

gsrohde:
@dlebauer See pull request #378 and notes there.

gsrohde:
Fixed by release 4.5.

Issue #355, opened 2015-10-01
Error when adding a new user to BETYdb


serbinsh:
I keep having this issue when I add a user to BETY.  I enter in all of the info and hit the sign up button, which then results in an error message (see attached)

![screen shot 2015-10-01 at 1 17 00 pm](https://cloud.githubusercontent.com/assets/3067369/10228087/012978ec-683f-11e5-96b1-a5aefb6b1894.png)

But if I check the users afterword the user is in the system and can log in (see testuser below)

![screen shot 2015-10-01 at 1 17 21 pm](https://cloud.githubusercontent.com/assets/3067369/10228106/16ad9c52-683f-11e5-9ce8-58303aff5864.png)

serbinsh:
@dlebauer Have you seen this before?  Is this an issue with our instance of BETY?

serbinsh:
@robkooper or @mdietze Have you had this issue before? I searched the issues but didn't see anything like this posted yet.

robkooper:
@serbinsh can you check in the log/production.log file?

gsrohde:
@serbinsh Was this on betydb.org or some other deployment?  If some other, what is the git version of the deployment?  I didn't see anything suspicious in the betydb.org log files.

dlebauer:
@serbinsh does this still occur?

dlebauer:
@gsrohde this just occurred when I signed up for ebi-forecast.igb.illinois.edu/bety-mepp

gsrohde:
Funny, I had no trouble.

dlebauer:
Perhaps try signing up with a request for elevated permissions

gsrohde:
@dlebauer It looks like a new user was made in spite of the error message.  And you must have gone in via psql and changed the access levels.  Yes?

I did try with elevated permissions too and still had no problem.

gsrohde:
I'll see if I can tell anything from the logs.

dlebauer:
Yes, the user was successfully created

I did manually change access_leve

Here is the log:

```
args = []; block =
  Rendered users/new.html.erb within layouts/application (41.1ms)
Completed 200 OK in 58.2ms (Views: 57.2ms | ActiveRecord: 0.0ms)
Started POST "/bety-mepp/users" for 128.174.125.237 at 2015-11-30 12:53:18 -0600
Processing by UsersController#create as HTML
[...]

Issue #368, opened 2015-11-17
Allow separate instances to customize UI elements 


dlebauer:
It would be great if users could have a simple text file that can be edited for customizing some elements of the user interface.

Ideally, we could a) use some conventions and b) coordinate this with customization of Clowder and interfaces (@robkooper?) and other tools within NDS (?)  

Key customizations are necessary; 'nice to have' customizations would be nice, if it is easy to do. From what I've seen in Jekyll, passing variables around isn't difficult. 

In Jekyll and some Ruby apps this is done via a local `config.yml` file. These seem very easy for non-ruby developers (e.g., me) to understand and modify. There are alternative ways to implement this feature but simplicity is important. (there is also a Ruby Gem http://www.rubydoc.info/gems/configatron/4.5.0 )

### Key customizations:

1. website title [`spec/features/static_integration_spec.rb`](https://github.com/PecanProject/bety/blob/master/spec/features/static_integration_spec.rb)
2. website description
1. data citation (can be blank)
2. contact information
3. git sha in footer and also version number
4. powered by betydb with link back in footer
<img width="412" alt="screen shot 2016-09-28 at 3 58 08 pm" src="https://cloud.githubusercontent.com/assets/464871/18932061/73192290-8594-11e6-8607-3f9f818f275d.png">

5. model output maps removed from all instances except betydb.org

### "Nice to have:

1. color scheme
1. sponsor logos
3. photograph on home page
4. descriptive paragraph
5. link to associated instance of PEcAn

![image](https://cloud.githubusercontent.com/assets/464871/11227002/0cfd6e42-8d49-11e5-8285-028966e16b9d.png)

### Misc updates (not on per-site customization, just while you are in there)

1. Move search bar on home page in line with the "Welcome to BETYdb", widen text. 
2. Change name of access level 2 from  'EBI' to 'Internal'

max-zilla:
Is there any concern about others branding this to the point it's unrecognizable? I assume there are some elements that must be present, e.g. BETYdb identification and such. We may want some required info that can't be removed or hidden.

dlebauer:
Do you think that a 'powered by BETYdb' along with links to the software
license and the github repository would suffice?

The aims are to facilitate scientific discovery and attribution.

max-zilla:
That would address my primary concern - making sure there's a path back to the project github at the very least. Even a clickable grayscale BETYdb logo being included where Illinois and Berkely are in the image above would be enough, I think.

[...]

PR #378, opened 2015-12-14
Fix for GH issue #377.

gsrohde:
@dlebauer This version is viewable at http://pecandev.igb.illinois.edu/sr.  If you are going to test it out, feel free to change the email of (e.g.) the carya user to your own.  It should only send e-mail if the user attempts to _increase_ their access level beyond the default access level.  Note that any user with `page_access_level = 1` can change their own (data) `access_level` to 1 as well, or for that matter, that of any other user.  I didn't change this.  This is the way it has been.

@max-zilla Here's more code for you to look at.


robkooper:
@gsrohde can you add a check to https://github.com/PecanProject/bety/blob/32a03c83b0f2e411e46aab2be6a70296a628ff7e/app/mailers/contact_mailer.rb#L62 to make sure that there is an email_to otherwise send an email to the from saying no admin is specified.

gsrohde:
@robkooper What is the use case for this?  Users who install their own copies and don't specify an administrative user?  I assume you want this in order to prevent the bug we were seeing in issue #355.

But I think there is a better way.  For one thing, we already have the assumption that there is an administrator at betydb@gmail.com.  This is the hard-coded sender (or at least reply-to address) when sending out API keys, and it is the contact address listed in the footer of every page.  So we could just always add it to the list of admins and then this list would never be empty and we wouldn't see bug #355.

But (per #368) I'm betting that we really don't want this address hard-coded.  In this case, we should make it a required set-up step to provide an administrative e-mail address just as we require a database configuration file and a site keys file.  Or perhaps the database-initialization/synchronization script can prompt for a admin user if there isn't one.

@dlebauer Do you have thoughts about this?

robkooper:
@gsrohde the betydb@gmail.com does not work for anybody but David, nobody that installs betydb will have access to that email. I think the best way is for the first user to ask for admin rights to be granted admin rights. Then if you remove admin from a user make sure that it is not the last user (i.e. you should never be able to remove admin from yourself). 

I think also email send from betydb should come from the first admin found. This way we can get rid of the betydb@gmail.com account.

Let me know how that sounds.

dlebauer:
> email send from betydb should come from the first admin found.

Sounds good except the contact email (and new user from email) should be
one of the configurable properties.

gsrohde:
@robkooper @dlebauer 

I've made the contact e-mail address and phone number shown on the home page configurable, and I added this e-mail address to the list of admins that get mailed so that the list is never empty.  This fixes bug #355.

If there are no serious objections, I'll pull request this and add notes about making a configuration file to the documentation.

dlebauer:
@gsrohde looks good. After merging this, you can close #377 and #355, correct?

Two questions:

1. Did you add information about use of the `customization.rb` template to the documentation?
2. would the `customization.rb` file be the place where all site-specific UI customizations can be made ( #368)?

config/customization.rb.template
```diff
@@ -0,0 +1,2 @@
+ADMIN_EMAIL = "betydb@gmail.com"
```
robkooper: Would it make sense to merge this with config/additional_environment.rb instead of having 2 files that need to be modified?

gsrohde: I considered this but decided against it for a few reasons.  One is that the settings in the two files seem largely orthogonal, and I liked having a clearly-named place to put customization settings with a mind to possible expanding the available settings to encompass all of what might be required to address issue #368.  A less good reason was that I wasn't that familiar with `additional_environment.rb` and wanted to start somewhere fresh.  Third, I was thinking this may only be a temporary solution to an immediate problem, and if I'm going to eventually jettison this infrastructure in favor of something more comprehensive, it's nice to have those settings all in one place.  If you feel strongly about combining the files, I don't have a tremendous problem with doing so.

dlebauer: > settings in the two files seem largely orthogonal

depends on if you are a splitter vs. a lumper. Currently `additional_environment.rb` has information about how to send emails and this `customization.rb` file has information about where to send the emails. So that doesn't seem orthogonal. More generally,

>  if I'm going to eventually jettison this infrastructure in favor of something more comprehensive,

I think that is what needs to be done. We can have a single file and instructions for each line documented, similar to what we do with [the `pecan.xml` settings file](https://github.com/PecanProject/pecan/wiki/PEcAn-Configuration). A more relevant example would be the way Jekyll uses a [`_config.yml`](https://mmistakes.github.io/so-simple-theme/theme-setup/#site-setup) file.

> A less good reason was that I wasn't that familiar with additional_environment.rb and wanted to start somewhere fresh. 

Luckily we have the author Rob here ...

gsrohde: One case where "splitting" is definitely preferable to "lumping" is if you are going to want multiple template files for each aspect since in general the number `n_1 x n_2 x ... x n_m` of template files required by lumping is greater than the `n_1 + n_2 + ... + n_m` template files required for splitting if `m` and the `n_i's` are all >= 2.  But since in our case we only have two `additional_environment` templates and only one `customization` template, we'd actually eliminate one template file by lumping.  (I _had_ considered making per-environment customization files, though, so that the "real" administrator wouldn't get e-mail from developers doing testing, but I tabled that idea for now.)  I still think the two sets of setting are largely orthogonal, however, since _where_ one sends an e-mail is likely to depend on different factors than _how_ one sends e-mail, and also because the latter seems more technical.  But I won't belabor this any further and will wrap the customizations into `additional_environment`.

I would eventually like to fold all the site and instance-dependent settings into a yml or xml file.  Ryan Bates has a Rails-cast on how to put config settings into a yml file.  But I'd probably prefer xml if we're going to want to put anything very complex in there, such as the HTML snippet to use for a footer.

gsrohde:
I forgot to mention one other reason I used `customization.rb`.  I read it in in a different context from the way `addtional_environment.rb` is read in, and I didn't want to have to go through the trouble of figuring out how to make the constants I set immediately available to the layout template and the mailer if they are set in `additional_environment.rb` instead.

gsrohde:
@robkooper 

I actually don't see any instance on `ebi-forecast` or `pecandev` where `additional_environment.rb` gets used.  Is this mainly for overriding defaults on a VM?  Where is the documentation about it?  (I think this is another (subliminal) reason why I didn't use it.)

gsrohde:
Here's what I've done:

I got rid of `customization.rb` and moved toward using a YAML file.  So far, there are only two settings that will be used (admin's contact phone and email).  I plan to get rid of `additional_environment.rb`:  Except for `config.default_user` and possibly the SMTP domain, all of the settings in `additional_environment_vm.rb` are the default values anyway, so there's really no need to set them.  As for `additional_environment_ebi.rb`, only the SMTP address and domain are not the default values, but we don't seem to be using this, even on `ebi-forecast`, so I'm guessing the defaults work just as well.  (I've  left the machinery that reads this file in place until we're sure what, if anything, we still need.)

I've deliberately kept the YAML file structure simple--only top level settings with optional per-environment sections.  If we decide later we need a more hierarchical structure, I will implement the machinery to deal with this when and if the need arises.  But just wanted to be able to move forward on this for now.

robkooper:
Why not move the SMTP/default.user settings to the YAML file as well. I'd prefer to have to point people only to a single file that needs to be modified.

robkooper:
Also can we make sure that the config/application.yml file is in .gitignore so people don't accidently check in their version.

Would it make sense to look at other configuration variables and collect them all in this one location (for example:
* config.time_zone = 'Central Time (US & Canada)'
* REST_AUTH_SITE_KEY = '....'
* REST_AUTH_DIGEST_STRETCHES = 10

gsrohde:
@robkooper 

I agree that everything that is likely to vary between deployment instances should go in the YAML file.  I was going to wait to do this until I started serious work on issue #368, but I guess I'll go ahead and add everything that we are already customizing.

Thanks for reminding me about the .gitignore entry; I had planned to do that.

As for the SMTP settings, the Rails defaults are
```
config.action_mailer.delivery_method = :smtp
config.action_mailer.smtp_settings = {
  :address => 'localhost',
  :domain => [might also be 'localhost'--not sure],
  :port => 25
}
config.action_mailer.perform_deliveries = true
config.action_mailer.raise_delivery_errors = true
config.action_mailer.default[:charset] = "UTF-8"
```
though the test and development environment configuration files override the defaults for `delivery_method` and `raise_delivery_errors`, respectively.
(Note that Rails 3.2 uses `config.action_mailer.default[:charset]` in place of `config.action_mailer.default_charset`.)
Is there any instance where we don't use the defaults?  Is there likely to be?  There are many other configuration options that we _could_ add to the YAML configuration file, but I think it makes sense to keep things simple and start with those options that are likely to actually get used to override default settings.

As for `config.default_user`, I grepped the Rails code base but couldn't find anywhere where it gets used.  What is it supposed to do?  I think I read something about what it does in the documentation somewhere, but I couldn't find that either.

robkooper:
I think the config.default_user is a leftover and be ignored (now is guestuser account). I think ebi-forecast had different setting for mailserver. I tend to use localhost and install postfix.

gsrohde:
Will now begin updating documentation.

PR #379, opened 2016-01-06
Gh 377 alt

gsrohde:
@dlebauer @robkooper See if you like this better.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -70,7 +70,7 @@ class UsersController < ApplicationController
   end
 
   def edit
-    # Allow admins to edit any account, and each user thier own
+    # Allow admins to edit any account, and each user their own
     if current_user.page_access_level == 1 or params[:id] == current_user.id.to_s
       user_id = params[:id]
     else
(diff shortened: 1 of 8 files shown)


dlebauer:
Currently, the two `additional_environment_[ebi,vm].rb` are not DRY. The files only differ in the following:

1. values of `address` and `domain`
2. the vm file has `config.default_user=1` not clear if why this is in one but not the other

It also doesn't provide users with a canonical default or template that contains all of the available configuration options, as is done with a `_config.yml` file.

To make it DRY, can we just put the values that no one is likely to change (e.g. do we really want `default_charset = "utf-8"` to be an option? Why can't it go in `environment.rb`?)

Why can't we have an `additional_environment.rb` like

```ruby
config.action_mailer.smtp_settings = {
  :address => 'express-smtp.cites.uiuc.edu',
  :domain => 'igb.uiuc.edu',
  :port => 25
}
::ADMIN_EMAIL = "betydb@gmail.com"
::ADMIN_PHONE = "(217) 300-0266"
```

Though looking forward to using a  a `_config.yml` for #368. 

```yml
admin: 
  email: betydb@gmail.com
  phone: "(217) 300-0266"
  twitter: BETYdatabase
smtp: 
  address: express-smtp.cites.uiuc.edu
  domain: igb.uiuc.edu
  port: 25
```

P.S. I don't think anything in the scope of #368 requires embedded HTML, but if we do need to there are many options (e.g.  [HTML and other arbitrary code snippets can be embedded in YAML](http://stackoverflow.com/a/1786044/199217))

gsrohde:
@dlebauer If you haven't already, please read my last three comments on https://github.com/PecanProject/bety/pull/378.  Another way of stating the first one is that "splitting" as opposed to "lumping" makes it easier to make sample files DRY.

Though I'm not sure how much the concept of DRY applies to sample files, only one of which will be copied into the used code base.  Also, the changes I proposed make these two files much less repetitive than they were before (they currently differ only in the SMTP address and domain and in the fact that the VM version includes a default user; the other three settings are identical).  Just to be clear, `additional_environment.rb` isn't in the repository; you have to supply it, and having a sample version makes it easier to do so.

dlebauer:
I did read your notes. I think the problem with sample versions is that
they don't necessarily contain all of the available fields. That is why I
was suggesting either a default or a template configuration file, and why I
used the example of the `_config.yml` files used in Jekyll as one that we
could follow.

I don't want to belabor the point, I just want to be able to make progress
on having a configuration file moving forward.

PR #383, opened 2016-02-05
Fixed some model associations.

gsrohde:
@robkooper Could you look this over and see if you have any concerns?  I had to add these to get some of the API calls to work.

Also, I seem to recall there is one table whose name makes it look like a join table but it really isn't.  Which one was it?


gsrohde:
@robkooper Please OK this (or not).  Or, if you'd rather, re-assign to @max-zilla to look over.

robkooper:
Looks good.

Issue #151, opened 2014-12-02
1. Get Assets Pipeline working


gsrohde:
Some steps:
1. Replace all hard-coded asset references (esp. image tags) in view templates.
2. Test out production setup on pecandev.
3. Start drafting instructions for users.

gsrohde:
Completed by release 5.0.

PR #388, opened 2016-03-01
Rails 3.2 with assets pipeline

gsrohde:
This pull request addresses issue #151—it is a long-overdue preliminary upgrade in preparation for upgrading to Rails 4.  (We're currently only on Rails 3.2, and Rails 5 is due to be released mid-month (a beta version is already out).  More import than "keeping up with the latest" for its own sake, as maintainer of the BETYdb Web app code, I am constantly running into situations where I think "This would be so much easier if I was using a later version of Rails!" or where I read a blog or tutorial on how to implement a feature and then find it assumes I'm using Rails 4.  Moreover, I'm not sure how much longer Rails 3.2 will continue to be maintained and supported.)

@dlebauer, @robkooper, @mdietze, @serbinsh, @max-zilla and any other interested users: I've deployed this code to the main pecandev deployment at http://pecandev.igb.illinois.edu/bety.  I'd appreciate if you take the code for a spin and see if you notice any problems.  

**_Better yet, try deploying your own copy of this version._**  (Although it should be easy enough to switch an existing copy and then revert to the master branch after you are done, if you prefer not to mess with your existing working copy you can simply `cp` your whole working copy to a new directory by doing this in the parent directory:

```
cp -R <current rails root directory> <new rails root directory>
```

Then `cd` to the new directory and do a `git fetch` followed by `git checkout --track origin/rails32-with-asset-pipeline`.
Alternatively, `git clone` this branch of `PecanProject/bety` to some location and copy the customization files from your current working copy.  Running
    `
    git status --ignored config
`
in the root directory of your old working copy will remind you what files might need to be copied.)

Once you have a working copy on the `rails32-with-asset-pipeline` branch, do the following to get it working:
1. `cd` to the root directory of the working copy (the "Rails root" for that copy).
2. Do
    `
    bundle install
`
to get required new Ruby gems.
3. Assuming you run your copy of BETYdb in production mode (probably the case if you run it under Apache), compile the assets with
    `
    bundle exec rake assets:precompile RAILS_RELATIVE_URL_ROOT=/bety
`
This assumes you have configured Apache to serve your BETYdb instance at `http://<your host>/bety`.  Change the directory name accordingly if you serve it at a different subdirectory; or, if you just serve it at `http://<your host>`, leave off the `RAILS_RELATIVE_URL_ROOT` setting.
4. If you have configured this copy to run under Apache, either restart Apache or just do
    `
    touch tmp/restart.txt
`
in your Rails root directory.  If this is a new copy, and you don't want to bother configuring Apache to run it, you can still run it in production mode by doing the command
    `
    rails server -eproduction
`
in the Rails root directory.  But first you must edit the file `config/environments/production.rb` and change the line
    `
    config.serve_static_assets = false
`
to
    `
    config.serve_static_assets = true
`
The app will be available at http://localhost:3000 on your local machine.

Please let me know of any problems with the code on this branch or with these instructions!


dlebauer:
I don't want to get too far into the weeds here, especially because there is a good chance my entire ruby installation is borked, but when I do `bundle install`, I get errors associated with the nokogiri gem. It says:

> An error occurred while installing nokogiri (1.6.0), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.6.0'` succeeds before bundling.

And if I do `gem install nokogiri -v '1.6.0'`, I get

> [bety]$ gem install nokogiri -v '1.6.0'
Building native extensions.  This could take a while...
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/ext/builder.rb:54: warning: Insecure world writable dir /Users/dlebauer in PATH, mode 040777
ERROR:  Error installing nokogiri:
	ERROR: Failed to build gem native extension.

see full output here: 
[output.txt](https://github.com/PecanProject/bety/files/216058/output.txt)

Perhaps this is a good use case for docker ...

gsrohde:
It looks like libxml2 is missing.  This may be part of the Xcode developer tools.  You could probably comment out that Gem and not miss out on too much functionality if you don’t want to try to get it installed.

dlebauer:
@gsrohde what do I need to comment out from the Gemfile? I commented out nokogiri but `bundle install` still tries (and fails) to install this gem. And I can't find the string `xml` in the Gemfile.

gsrohde:
@dlebauer In addition to commenting out nokogiri, you'll have to comment out capybara and capybara-webkit too unless you used the '--without' flag.  E.g.
```
bundle install --without=test javascript_testing
```
This is because capybara depends on nokogiri, and capybara-webkit in turn depends on capybara.  You can see the dependency chain in Gemfile.lock.

Alternatively, you could install libxml2 on your machine.  Assuming it's a Mac, this may be available through XCode.  Or you could use homebrew or MacPorts.  libxml2 comes pretty standard on Linux machines, I think.

dlebauer:
libxml2 is installed

```
[bety]$ brew install libxml2
Warning: libxml2-2.9.2 already installed
```
I think I have Xcode installed. 

And I get another error.

>An error occurred while installing pg (0.17.0), and Bundler cannot continue.
Make sure that `gem install pg -v '0.17.0'` succeeds before bundling.

gsrohde:
Do you have a PostgreSQL installation on your machine?

dlebauer:
@gsrohde what is the status of this pull request - does it contain important features that should be merged?

gsrohde:
This is a long-standing task I've wanted to get done but one that always seemed to have lower priority than other issues.  I expect at this point it might take a few days of work to work out all the kinks involved in upgrading.  Releasing a new upgrade should probably wait until after any demo you may be doing next month, but I could do preliminary preparatory work before then.

robkooper:
keep in mind that in December there is AGU and we should not do this before AGU. So probably the beginning of January is a good time for this.

gsrohde:
Superseded by PR #577.  Closing.

PR #390, opened 2016-03-04
Change numeric types to floats

gsrohde:
Changed data type of columns representing measurements or statistical values to floating point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,57 @@
+# use ruby
+language: ruby
+
+# use containers
+sudo: false
+
+# versions of ruby to use
+# see the list at http://rubies.travis-ci.org/
+rvm:
+  - 1.9.3
+  - 2.0.0
+  - 2.1.8
+
+# used for all ruby tasks to specify the database
+env:
+  - RAILS_ENV=test
+
+# cache packages and bundler
+cache:
+  - packages
+  - bundler
+
+# install R: use r-packages-precise (https://cran.r-project.org/bin/linux/ubuntu/precise/) 
+# as source which is white listed (https://github.com/travis-ci/apt-source-whitelist/)
+addons:
+  postgresql: 9.4
+  apt:
+    packages:
+      - postgresql-9.4-postgis-2.1
+      - postgresql-9.4-postgis-2.1-scripts
+  
+# notifications should go to gitter
+notifications:
+  webhooks:
+    urls:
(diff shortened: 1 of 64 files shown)


gsrohde:
@robkooper Could you look this over?

dlebauer:
@gsrohde is there a reason to assign this to @robkooper instead of @max-zilla? Max has more time on the project, so by default it would more sense to ask him to review.

robkooper:
Also did you test this with pecan yet?

gsrohde:
@dlebauer I assigned to Rob because he is more familiar with the database, how it is used, and how this change might affect current users (if at all).  But of course, ideally Max will become familiar enough with BETYdb and how it is used to be able to evaluate this.  @robkooper Please go ahead and assign to Max if you deem appropriate.  Also, @robkooper: I didn't test with Pecan yet because I'm not familiar with how to do this.

dlebauer:
To test with pecan

1. Download VM https://opensource.ncsa.illinois.edu/projects/artifacts.php?key=PECAN
2. Import into VirtualBox
3. Launch PEcAn VM 
4. login is carya password illinois
5. (to confirm there are no errors on the fresh machine) run `/home/carya/pecan/scripts/build.sh --check --tests`
5. deploy branch of bety with these changes
6. run `/home/carya/pecan/scripts/build.sh --check --tests`
7. If all goes well you are done!

max-zilla:
Is still a good idea for @robkooper to look, but permit me to at least confirm I understand the gist (not much experience with ruby so mostly from a generic RDBMS perspective)...

- **self.up()** fires when you migrate the db to update the datatypes, and **self.down** is the rollback code to undo that migration. down is also called if invalid statement is encountered during the table drop/create steps of the migration.

- **up()** drops 4 views... really 2 primary views, one that is the union of those two, and a fourth that filters only to [checked >= 0]. it then changes 10 fields across 5 tables to float data type instead of numeric. it then recreates the views.

* **down()** does the opposite.

- in **simple_search.rb** a search handler was implemented that uses a regex to check for recognized digits/decimals/negative signs in the query when the queried column has float type.

I don't understand why the views are dropped and recreated. It doesn't look like any of the views definitions change and I didn't see anything obvious about that being necessary for change_column on the [Api reference](http://api.rubyonrails.org/classes/ActiveRecord/Migration.html). Why is dropping and recreating necessary, is there some indexing or requirement stuff behind the scenes?

gsrohde:
@max-zilla I pretty sure that you have to drop the views before changing any tables they depend upon.  This is not a Rails requirement but rather a PostgreSQL requirement.  It's a pain, but necessary I think.  I'd love to find out I'm wrong.  You can verify by taking the view stuff out of the migration and seeing if it works.

max-zilla:
@gsrohde bingo, you are correct:
```
bety=> ALTER TABLE traits ALTER COLUMN mean TYPE float;
ERROR:  cannot alter type of a column used by a view or rule
DETAIL:  rule _RETURN on view traitsview_private depends on column "mean"
```

By my estimation your script looks good otherwise. Also tested out the regex and successfully identified whole/decimal numbers. It doesn't catch negative values, if it's possible to encounter those - seems plausible in one of the **parama** **b** **c** fields? If so, this small adjustment should do it:

```
/(-?(\d+(\.\d*)?)|(\.\d+))([eE][+-]?\d\d?)?/
```
(just added "-?" after first parentheses)

gsrohde:
@max-zilla Good catch—I've added the `-?.`.

gsrohde:
I notice one remaining issue: With the current data in the database, the migration is not reversible.  (This is perhaps not a big issue—ideally, we'll never have to reverse the migration.)

The problem is that there are four `covariates.level` values of 999999999999.9999.  The up migration converts this to 10^12.  This is too large to get converted back to type NUMERIC(16,4), so the down migration fails.

@dlebauer Could you verify that these values are correct?  The covariate id numbers are 11199–11202.  Actually, we could just replace these values with 999999999999.9990 because Postgres floats can keep 15 significant digits and so won't round this to 10^12.

Note that I _had_ thought I was converting to single precision floats (PostgreSQL _reals_) which have at least six decimal digits of precision—plenty for this database, I thought.  But it seems that the default for floats is double precision, and to make the migration use single precision I'd have to use raw SQL in the migration.  Since double precision floats use 8 bytes of storage, that's still less than what NUMERIC(16,4) uses.  So I think I'll keep the migration as is.

dlebauer:
On a new PEcAn VM v 1.4.5 I ran `scripts/build.sh --check --tests`

This test gets all the way until ED catches a segfault (pecanproject/pecan#746). Everything else (including package tests) works. (Although I _could_ install v 1.4.4 and run the tests, 95% of the pecan functionality has been tested against this change).

So I give :+1: its up to @robkooper to merge

robkooper:
does the test pass without the change or does it fail as well?

dlebauer:
@gsrohde can you put a dump of the updated database that I can check against the PEcAn tests? I don't think it is necessary to worry about changes to the front end / migrations, and if you can just provide a dump it will be easy to import and run tests against.

gsrohde:
@dlebauer I created a copy of `ebi_production` on `ebi-forecast` and then ran the migration on it.  It's called `ebi_production_proposed`.  If you actually just want a dump of it, run `pg_dump ebi_production_proposed > some_file.psql`, or let me know if you want me to.

gsrohde:
David: I’m planning a new release today.  I know you’re probably busy, but if you think you’ll get this tested today, I will include the migration in the release. —Scott

dlebauer:
I think it is okay to move forward with the migration. All of the PEcAn tests passed until there was an unrelated bug with the ED2 integration test on the VM (PecanProject/pecan#746). However, this is the last test and is unrelated to changes in the database.

@robkooper is it okay if Scott adds this migration?

robkooper:
Sure.

dlebauer:
@gsrohde go ahead and deploy this today.

PR #394, opened 2016-03-15
Add Dockerfile & docker-compose.yml for BetyDB

max-zilla:
To use, cd to the bety directory with docker installed and simply say:
```
docker-compose up
```

This will create a container with a fresh PostGIS install and then create a clean BetyDB container to connect to it. Once the compose script has completed you should see something like this:
```
bety_1     | [2016-03-15 19:36:44] INFO  WEBrick 1.3.1
bety_1     | [2016-03-15 19:36:44] INFO  ruby 2.1.8 (2015-12-16) [x86_64-linux]
bety_1     | [2016-03-15 19:36:44] INFO  WEBrick::HTTPServer#start: pid=104 port=3000
```

This means Bety is ready. In a new terminal use 
```
docker ps -a
```
...to see your running docker containers. There will be a port number listed in the line with bety, e.g.
```
0.0.0.0:32789->3000/tcp
```

You can access bety at 192.168.99.100/port locally, e.g. in this example

```
192.168.99.100/32789
```

PR #395, opened 2016-03-16
Added docker files

max-zilla:
Add Dockerfile and dependencies to Bety. This is a new pull request to replace https://github.com/PecanProject/bety/pull/394 after correcting some git issues.


Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ruby:2.1
+MAINTAINER Max Burnette <mburnet2@illinois.edu>
+
+EXPOSE 3000
+
+# Install dependencies
+RUN apt-get update &&\
+    apt-get install -y sudo curl unzip netcat &&\
```
robkooper: do we still need unzip?

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add - &&\
+    apt-get -y update &&\
+    apt-get -y upgrade &&\
+    apt-get install -y postgresql-client-9.4
+
+COPY / /home/bety
+
+RUN cd /home/bety &&\
+    /bin/sed -i "/capybara-webkit/ s/^/# /" Gemfile &&\
+    gem install bundler &&\
```
robkooper: I think you can use  `gem install bundler  --without javascript_testing` instead of the sed command.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    bundle install
```
robkooper: I think you can use  `bundle install --without javascript_testing` instead of the sed command.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+  ports:
+    - "3000"
+  links:
+    - postgres:pg
+
+# postgres database
+postgres:
+  image: mdillon/postgis:9.4
+  ports:
+    - "5432:5432"
```
robkooper: remove the :5432 lets leave it random by default.

docker/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
...
+        echo "Error: Timeout wating for Postgres to start"
+        exit 1
+      fi
+    done
+
+    # Move database config file into config directory & add host/port
+    cd /home/bety/config
+    /bin/cp /home/bety/docker/database.yml database.yml
+    /bin/sed -i "/host:/ s/$/ $PG_PORT_5432_TCP_ADDR/" database.yml
+    /bin/sed -i "/port:/ s/$/ $PG_PORT_5432_TCP_PORT/" database.yml
```
gsrohde: Are these two lines really doing what you want them to do?  My sed version says the -i option is for providing an extension for renaming the unmodified version of the file.  Also, where do `PG_PORT_5432_TCP_ADDR` and `PG_PORT_5432_TCP_PORT` get set?  Is this a docker thing?

robkooper: sed -i will create a backup and replace the file you are working on. If not given an argument it will just replace the original file. On a mac this does not work however and you will need to specify an argument for -i. The PG_PORT are indeed environment variables automatically set by docker.

docker/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
...
+    if ! psql -h $PG_PORT_5432_TCP_ADDR -p $PG_PORT_5432_TCP_PORT -U postgres -lqt | cut -d \| -f 1 | grep -w "bety" > /dev/null ; then
+      echo "Creating bety database"
+      createdb -U postgres -h $PG_PORT_5432_TCP_ADDR -p $PG_PORT_5432_TCP_PORT bety
+    fi
+
+    # Download & initialize Bety database contents
+    cd ../script
+    curl -LOs https://raw.githubusercontent.com/PecanProject/pecan/master/scripts/load.bety.sh
+    chmod +x load.bety.sh
+    ./load.bety.sh -a "postgres" -p "-h $PG_PORT_5432_TCP_ADDR -p $PG_PORT_5432_TCP_PORT" -o postgres -c -u -g -r 0 -m 99
```
gsrohde: I'm making a rake task to populate the database (probably replacing update-betydb.sh).  It gets most of these argument values from the database.yml file.  Once I check it in, you can replace the previous three lines with `bundle exec rake bety:db:populate`.

docker/entrypoint.sh
```diff
@@ -0,0 +1,39 @@
...
+    # Move database config file into config directory & add host/port
+    cd /home/bety/config
+    /bin/cp /home/bety/docker/database.yml database.yml
+    /bin/sed -i "/host:/ s/$/ $PG_PORT_5432_TCP_ADDR/" database.yml
+    /bin/sed -i "/port:/ s/$/ $PG_PORT_5432_TCP_PORT/" database.yml
+
+    # Create bety database if it does not exist
+    if ! psql -h $PG_PORT_5432_TCP_ADDR -p $PG_PORT_5432_TCP_PORT -U postgres -lqt | cut -d \| -f 1 | grep -w "bety" > /dev/null ; then
+      echo "Creating bety database"
+      createdb -U postgres -h $PG_PORT_5432_TCP_ADDR -p $PG_PORT_5432_TCP_PORT bety
```
gsrohde: @robkooper Is there any reason the `-c` (create database) option to `load.bety.sh` doesn't actually create the database?  It seems to only drop and recreate the schema.

robkooper: It actually does not create the database. It just drops the default schema and readds it.

gsrohde:
@max-zilla Any reason you put some files in the Rails root directory and others in the docker/ directory?

max-zilla:
This was largely due to the way docker creates containers. By putting the Dockerfile at the root, docker is prepared to copy the contents of the root directory (i.e. all the Rails app code) into the docker container rather than, say, using curl or git to pull them down inside the container from the web.

The docker-compose file could be moved into the docker directory I suppose. The other files I put in a /docker directory to avoid cluttering up the root any more than necessary, but I believe docker-compose can be moved without any issue. Dockerfile should stay in root however.

PR #397, opened 2016-03-17
travis build

robkooper:
This will enable travis to auto build bety.


robkooper:
@dlebauer @max-zilla please check

dlebauer:
does it load data at any point, or just create an empty database?

robkooper:
create empty test database, no data loaded or needed to be loaded.

PR #405, opened 2016-04-27
Merge remote-tracking branch 'PecanProject/master'

robkooper:


robkooper:
See https://github.com/PecanProject/pecan/issues/629 and https://github.com/PecanProject/pecan/pull/806

gsrohde:
@robkooper I don't quite get what this is doing.  Why does File changed show nothing?

robkooper:
Ok this time it should be good to go.

dlebauer:
Echoing Mike's comment on the UI the term 'clusters' is ambiguous. sitegroups isn't great but it is more clear.

The field 'everybody' also seems like an awkward term ... wouldn't 'public' be more straightforward? And is there a need to hide site groupings? (Is this just to keep the ad hoc groupings hidden?)

robkooper:
Can't use public since it is a reserved word in ruby. I didn't like network since that has a different conation for me. It needs to be something that fits in ruby land, which is why i picked cluster, so we can have  table called clusters_sites, so something with sitegroups has sites duplicated in the title.

gsrohde:
How about 'openaccess' (or even 'public_access') instead of 'everybody'.  Sure it's longer, but at least it's clear at a glance what it means.  Likewise, using 'sitegroups' in place of 'clusters' results in the redundant 'sites_sitegroups', but so what?  A little redundancy for the sake of clarity doesn't seem a bad trade.

mdietze:
I agree with @gsrohde and @dlebauer suggestions, both on preferring openaccess or public_access to everyone and on preferring sitegroups (or maybe sitenetworks) to clusters.  This fix needs to go in the next PEcAn VM, which we're hoping to build early this week, so hopefully we can get this sorted out, do a quick find-replace, and merge it in.

robkooper:
This is now continued at #407

robkooper:
renamed cluster to be sitegroup and everybody to be public_access

PR #407, opened 2016-05-02
Site Groups pecan 629

robkooper:
This contains the code to add a migration and interface to add site groups. This is a group of sites that logically belong together.

PR #408, opened 2016-05-02
Robkooper cluster pecan 629 modified

gsrohde:
Comments on pull request

* `Cluster#to_s` returned `hostname`, which throws a method missing error.  I made it return `name`.
* I made several changes to the migration.
 * I added three foreign-key constraints.
 * I made "everybody" non null.
 * I gave all timestamp column a default of utc_now() and these defaults are now used by the insert statements.
* Changes to the AuthenticatedSystem module:
 * I removed references to ClustersSitesController.  These controllers don't exist and so this list items have no effect.
 * I added "ClustersController.ALL" to the admin_requirement list.  This isn't really necessary since you added the fallback to level 1 if the controller action is not found in any list, but I think the fallback option shouldn't be used except as a stopgap to programming neglect.
 * I modified the Clusters index view template so that the "New Cluster" buttons is centered to be consistent with other pages.
 * I added some RSpec tests having to do with authorization to do various things with clusters.

Questions

* Are you really wanting to start the sequences at (1e9 * n) and not (1e9 * n) + 1?
* I assume that you have reasons unrelated to clusters for adding methods #hostid, #host_start, and #host_end to Machines.  But since we have #hostid, I used it in the migration in the two setval calls to avoid repeating code.  But if there's some reason not to, change it back.

### DIFF
--- a/app/models/cluster.rb
+++ b/app/models/cluster.rb
@@ -27,7 +27,7 @@ class Cluster < ActiveRecord::Base
   end
 
   def to_s
-    hostname
+    name
   end
 
 end
--- a/app/views/clusters/index.html.erb
+++ b/app/views/clusters/index.html.erb
@@ -5,7 +5,9 @@
 				<h1>Clusters</h1>
 			</header>
       <div id="process_container">
-        <div id="back"><%= link_to 'New Cluster', new_cluster_path, :class => "button button-primary" %></div>
+        <div id="back"><br /></div>
+        <div id="new"><%= link_to 'New Cluster', new_cluster_path, :class => "button button-primary" %></div>
+        <div id="forward"><br /></div>
       </div>
       <br />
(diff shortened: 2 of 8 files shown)


gsrohde:
@robkooper This is a pull request for merging into (a copy of) your original pull request.

The failing test has to do with the permissions problem I mentioned (I think).

gsrohde:
@robkooper On second thought, the failing build is, I think, caused by the fact that we don't update `production_structure.sql` until we do a new deployment.  This is because we want this file to reflect _exactly_ what's on `ebi-forecast` in the `ebi_production` database (I generate it there upon deployment and then check it in).  So I think Travis needs to call `rake db:migrate` after calling `rake db:structure:load` but before `rake db:fixtures:load`.  This problem came to light because there was a new migration _and_ a new fixture file that depended on it.

robkooper:
See #407 which incorporates these changes as well as renaming cluster to be sitegroup and everyboyd to be public_access.

PR #427, opened 2016-07-09
fix for rgeo

robkooper:
This is the fix to make rgeo to work correctly.


gsrohde:
Added an RSpec test to test RGeo installation.

PR #428, opened 2016-07-11
migration for benchmark code

robkooper:
See https://github.com/PecanProject/pecan/pull/895


robkooper:
@bcow This is your migration (small tweaks to it). Do you mind removing it from your pull request.

gsrohde:
@robkooper I made a few stylistic changes to the migration, but otherwise it looks fine.  The main practical effect of my changes (which may be seen by comparing the generated structure files) is that the timestamp columns (`created_at` and `updated_at`) now must be non-null.  The main question I have is that I was wondering if you intended to leave out the timestamp columns for most of the join tables—you only added them for some, not all.

Issue #411, opened 2016-05-06
can't add sites to site-group


robkooper:
On the VM I can not add a new site to AmeriFlux

serbinsh:
@robkooper I can't any new sites to a site group.  I am trying to create NGEE-Arctic and NGEE-Tropics but when I click the + nothing happens.  This is universal in the Site_Groups table.  @mdietze has had the same issue.   Can we get this bug fixed?

serbinsh:
![screen shot 2016-09-23 at 11 47 05 am](https://cloud.githubusercontent.com/assets/3067369/18792180/7972e5c8-8183-11e6-99ef-bde1233d89d8.png)

gsrohde:
Need to release.

PR #440, opened 2016-09-23
Fixes issue with not being able to add new sites to sitegroup #411

robkooper:


serbinsh:
@gsrohde I assume we just need to pull these updates (when merged) and restart BETY to get the fix, correct?

serbinsh:
@robkooper Can we merge this change so i can pull?

PR #442, opened 2016-10-03
Add experiments tables

gsrohde:
@robkooper, @dlebauer Please check that this looks OK.  See comments below.


gsrohde:
Re-doing pull request.

PR #443, opened 2016-10-03
Adding migration for creating experiments

gsrohde:
@robkooper, @dlebauer Please check that this looks OK. See comments below.

### DIFF
--- a/db/migrate/20160930213737_add_experiments_table.rb
+++ b/db/migrate/20160930213737_add_experiments_table.rb
@@ -6,6 +6,7 @@ class AddExperimentsTable < ActiveRecord::Migration
       t.date :end_date
       t.text :description, null: false, default: ''
       t.text :design, null: false, default: ''
+      t.integer :user_id, :limit => 8, :null => false
       t.timestamps
     end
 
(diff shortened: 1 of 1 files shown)


gsrohde:
@dlebauer and @robkooper: I added timestamp columns (`created_at`, `updated_at`) to all the tables.  (All existing tables have them except `trait_covariate_associations` and some of the benchmarks join tables.)

I added all obvious foreign-key constraints to the join tables.  What isn't as obvious is what the behavior should be if one tries to delete or update the id of a referred-to row.  One alternative is just to forbid this until the referring row is deleted.  Another is to automatically delete the referring row (in the case of deletions) or update the reference (in the case of updates of `id`).  I chose this behavior.

It just now occurs to me that experiment names should probably be unique.  If you agree, I'll add this constraint.  This and all the other textual columns are set to be non-NULL, and the others have a default value of the empty string.

The `start_date` and `end_date` columns are currently only constrained to be in the proper sequence (if they are non-null).  Let me know if other constraints on them are appropriate.

db/migrate/20160930213737_add_experiments_table.rb
```diff
@@ -0,0 +1,59 @@
+class AddExperimentsTable < ActiveRecord::Migration
+  def up
+    create_table :experiments do |t|
+      t.string :name, null: false
+      t.date :start_date
+      t.date :end_date
+      t.text :description, null: false, default: ''
+      t.text :design, null: false, default: ''
+      t.timestamps
```
robkooper: should also store the user if possible that created the experiment

robkooper:
Can we add at least an show page, if not also an edit page to the BETY interface.

gsrohde:
@robkooper I added the `user_id` column to the experiments table.  I'll add show and edit pages later.

Issue #444, opened 2016-10-03
Add a "site containment" table.


gsrohde:
This will be a table with two columns—call them, say, "container" and "member"—each referring to an id in the `sites` table.  It will have a row for each pair of sites `A` and `B` such that site `A`  completely contains site `B`, that is `ST_Contains(A.geometry, B.geometry)`.  This table will essentially serve as a cache of this relationship.  (How the table is populated and kept up to date remains to be determined.)

dlebauer:
closed because functionality handled by experiments and sitegroups tables

PR #445, opened 2016-10-03
Added table for caching the site-containment relation.

gsrohde:
@dlebauer Is this what you had in mind?

@robkooper Could you check this over?

See issue #444 for comments.

### DIFF
--- a/db/migrate/20161003171651_add_site_containment_table.rb
+++ b/db/migrate/20161003171651_add_site_containment_table.rb
@@ -1,0 +1,22 @@
+class AddSiteContainmentTable < ActiveRecord::Migration
+  def up
+    create_table :site_containment do |t|
+      t.integer :containing_site_id, limit: 8, null: false
+      t.integer :contained_site_id, limit: 8, null: false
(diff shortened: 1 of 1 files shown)


dlebauer:
Would it be sufficient to add a parent_id to the sites table?

```
site_id
parent_id
```

we may also want to have a field that describes the level of each site_id, i.e. as plant, plot, field, block, etc.

gsrohde:
I don't see any advantage, and generally it's a bad idea to be adding columns that will usually have nothing in them (with the possible exception of descriptive textual columns).  Moreover, for a site that is included in more than one other site, it's not clear which site `parent_id` should point to.

robkooper:
Lets table this for now. I think we can do a really good job with the following algorithm:
1) get geometry from site
2) find all data where site intersects with geometry provided (i.e. geometry from 1)

This allows us later to add an endpoint to BETY where we can just provide an arbitrary geometry and find all data inside that geometry. This is exactly what postgres is good at (might need to add a spatial index).

Issue #435, opened 2016-09-17
benchmarks_ensebles_scores` is missing a column


robkooper:
from @bcow : I just found a problem in bety that I could have sworn we already fixed ` benchmarks_ensebles_scores` is missing a column: it needs `t.text :score, :null => false`

robkooper:
Following are the migrations (in sql)
```
ALTER TABLE "benchmarks_ensembles_scores" ADD COLUMN "score" text NOT NULL;
INSERT INTO "schema_migrations" ("version") VALUES ('20161005181021');
```

gsrohde:
Fixed by release 4.13.

PR #446, opened 2016-10-05
Add missing column #435

robkooper:
@bcow this should add the missing column

### DIFF
--- a/db/migrate/20161005181021_add_score_benchmarks_ensembles_scores.rb
+++ b/db/migrate/20161005181021_add_score_benchmarks_ensembles_scores.rb
@@ -1,0 +1,9 @@
+class AddScoreBenchmarksEnsemblesScores < ActiveRecord::Migration
+  def up
+    add_column :benchmarks_ensembles_scores, :score, :text,  :null => false
+  end
+
+  def down
+    remove_column :benchmarks_ensembles_scores, :score
+  end
+end


db/migrate/20161005181021_add_score_benchmarks_ensembles_scores.rb
```diff
@@ -0,0 +1,9 @@
+class AddScoreBenchmarksEnsemblesScores < ActiveRecord::Migration
+  def up
+    add_column :benchmarks_ensembles_scores, :score, :text,  :null => false
```
gsrohde: @robkooper This really should be textual and not numeric?

robkooper:
This is what @bcow said to me.

db/migrate/20161005181021_add_score_benchmarks_ensembles_scores.rb
bcow: Yes - many "scores" will not be numeric

dlebauer:
Betsy can you give examples?

gsrohde:
@dlebauer Did Betsy ever answer this to your satisfaction?

dlebauer:
@gsrohde no, I still don't understand what scores would be non-numeric, but you can go ahead and merge / release this.

bcow:
@dlebauer Sorry I missed your response - this is part of @mdietze's original plan. Not all benchmarking functions will produce number scores - for example, functions that produce figures will have the path to the figure file in the scores field. Another example is a test that produces ordinal / categorical results.

Issue #438, opened 2016-09-22
Add cultivars.name, methods.name, and entities.name fields to views and search


dlebauer:
* [ ] Add a new field to traits_and_yields_view 
  * should be called 'cultivar' and should contain value in `cultivars.name`
* [ ] allow users to search on this field via web and API

gsrohde:
I have a draft fix in pull request #448.

gsrohde:
Implemented by release 4.13.

PR #448, opened 2016-10-13
Draft of changes for Github issue #438.

gsrohde:


app/models/traits_and_yields_view.rb
```diff
@@ -26,7 +26,7 @@ class TraitsAndYieldsView < ActiveRecord::Base
   SEARCH_INCLUDES = %w{ }
   SEARCH_FIELDS = %w{ scientificname commonname trait
                       trait_description city sitename author
-                      citation_year }
+                      citation_year cultivar entity cultivation_method }
```
dlebauer: what is `cultivation_method`?

gsrohde:
If you have a better suggestion for what name to use to refer the the method name, let me know.  I can't use `method` because this name conflicts with a built-in name Ruby uses.  Perhaps `method_name`?

I'll remove this from the searched-columns list, whatever we end up calling it.

The API search returns all columns of the view.

dlebauer:
`method_name` works

Issue #463, opened 2016-11-22
schema change for likelihoods for more precision


istfer:
Related to issue #462 

The SS values I calculate are very close to each other, such as `-35.38833`, `-35.38797` which all become `-35` in the database and can only be retrieved back as `-35`. 

The current table doesn't allow any decimal points for loglikelihood (`numeric(10,0) `):
```
                                       Table "public.likelihoods"
    Column     |              Type              |                        Modifiers                         
---------------+--------------------------------+----------------------------------------------------------
 id            | bigint                         | not null default nextval('likelihoods_id_seq'::regclass)
 run_id        | bigint                         | not null
 variable_id   | bigint                         | not null
 input_id      | bigint                         | not null
 loglikelihood | numeric(10,0)                  | 
 n_eff         | numeric(10,0)                  | 
 weight        | numeric(10,0)                  | 
 residual      | numeric(10,0)                  | 
 created_at    | timestamp(6) without time zone | default utc_now()
 updated_at    | timestamp(6) without time zone | default utc_now()
Indexes:
    "likelihoods_pkey" PRIMARY KEY, btree (id)
    "unique_run_variable_input_combination" UNIQUE CONSTRAINT, btree (run_id, variable_id, input_id)
    "index_likelihoods_on_input_id" btree (input_id)
    "index_likelihoods_on_run_id" btree (run_id)
    "index_likelihoods_on_variable_id" btree (variable_id)
Foreign-key constraints:
    "fk_likelihoods_inputs_1" FOREIGN KEY (input_id) REFERENCES inputs(id)
    "fk_likelihoods_runs_1" FOREIGN KEY (run_id) REFERENCES runs(id)
    "fk_likelihoods_variables_1" FOREIGN KEY (variable_id) REFERENCES variables(id)
Triggers:
    update_likelihoods_timestamp BEFORE UPDATE ON likelihoods FOR EACH ROW EXECUTE PROCEDURE update_timestamp()
```

I need to do more tests for other variables but at least for the `loglikelihood` column I would need 5 digits after the decimal

gsrohde:
@dlebauer, @robkooper Any reason not to use floating point types (e.g., double precisions) for these?  We changed most other numerics to floats in this migration: https://github.com/PecanProject/bety/blob/master/db/migrate/20160303221049_change_numerics_to_floats.rb

robkooper:
I see no reason not to.

gsrohde:
Fixed by release 4.14.

Issue #462, opened 2016-11-22
Need "unique_run_variable_input_combination" constraint to be removed


istfer:
I am working on a new implementation of the emulator method in PEcAn. For a given run, I compare  model outputs and data, and then calculate the sufficient statistics (SS) more than once for each proposed bias parameter. 

I want to store these (multiple) SS values (per unique run_id, variable_id and input_id) in the `likelihood` table but the `unique_run_variable_input_combination` constraint doesn't let me.

Is it possible to remove or relax this constraint for the likelihood table?

gsrohde:
@dlebauer Would you take a look at this request?  Were we rash in imposing this constraint, or is there a better alternative to the way she wishes to use this table?  I don't think I have enough domain knowledge to evaluate whether or not this is a reasonable request.

dlebauer:
@gsrohde we implemented these before they were used very much, and supporting an actual use case is a reasonable request. I'll punt to @robkooper and @mdietze

mdietze:
This is a reasonable request. Istem's the one who's been deep in the weeds on this part of pecan and I've discussed this with her -- relaxing this constraint is the best solution to the problem of needing to estimate non-model parameters during data assimilation. Specifically, for a given run/data/variable combination we need to evaluate multiple possible non-model parameters ( bias, variance etc). We want to store these evaluations in the DB since this is key to our plan for distributed data assimilation

robkooper:
I'd say for now drop the constraint.

gsrohde:
Fixed by release 4.14.

PR #464, opened 2016-11-29
New migration to address issues #462 and #463.

gsrohde:
@robkooper Could you check this over and see if it looks ok?


dlebauer:
@gsrohde I noticed that @robkooper gave thumbs up but am not sure if it pinged you

PR #447, opened 2016-10-10
Customization

gsrohde:
Here's a draft of an extension of the customization mechanism to most of the elements of interest on the Home page.  I suggest checking out this branch and trying out some of the features, either by creating or changing your `config/application.yml` file or temporarily tweaking the checked-in `config/defaults.yml` file.  I've deployed this branch to pecandev/beta; feel free to tweak that code also.  You will have to restart the Rails server (`touch tmp/restart.txt`) after each change you make in order for the results to be visible.  (If you are just running your own copy under the WEBrick Rails server, you just have to CTRL-C and reissue the command `rails server` or `rails s`.)

# Notes

I’ve implemented customization of almost all the outlined regions in the diagram shown in issue https://github.com/PecanProject/bety/issues/368.

image

Here are the keys for each item in the diagram, shown in left-to-right order starting at the top of the page and moving to the bottom:

```
organization:
    name:
    url:
    logo_file:
```
The “organization” key controls the link at the top left: “name” is the displayed text, “logo_file” is the name of the image file to be used (which should be present in /public/images), and “url” is the link target.  Note that in contrast to the diagram, the “BETYdb” portion of the heading is controlled by the next key.

```
site_identification_markup:
```
This takes a free-form value which allows any well-formed HTML markup.  It is the responsibility of the customizer to ensure that the markup works well within the general page layout.

```
homepage_heading:
```
This is just text (currently, "Welcome to BETYdb”).

```
homepage_body:
    lead_text:
    marked_up_block_text:
    photo:
        file:
        alt_text:
```
“lead_text” is just text; it cannot include markup.

“marked_up_block_text" is also text but may include markup.  (The default value includes a `<strong>` tag.)

"photo/file” should be the file name of the photo to use on the home page.  The file must be present in /public/images.

“photo/alt_text” is the text to be shown if the photo file is missing or the user has image display turned off.

#### <b>Top-left footer item (BETYdb)</b>

I have not yet implemented customization of this item.  Should it be linked somehow with the site_identification_markup (`"<strong>BETY</strong>db <small><strong>Biofuel Ecophysiological Traits and Yields </strong>Database</small>”`) or the homepage_heading (`"Welcome to BETYdb”`) or should it be completely independent for maximum flexibility?  I _did_ put the Maps & Data link shown underneath this item under the control of the configuration item that controls the display of the crop maps.

#### <b>Contact items</b>

These were handled previously.  The keys to use are “admin_phone” and “admin_email”.  Should I make the Gitter link configurable as well?

[...]

PR #467, opened 2016-12-07
updated application.template.yml

dlebauer:
I think it is useful to have all of the fields available in the template. I can't recall the rationale for not putting this information in the template, but in practice the user would need to open defaults.yml to determine what is configurable

Before merging this, please 

* [ ] add recommended image sizes
* [ ] update instructions for customization in the comments
* [ ] ~~also [update the documentation](https://www.gitbook.com/book/pecan/betydb-documentation/edit#/edit/master/distributed_betydb.md) but note that this should not repeat what is in the template (e.g. does not need a comprehensive list of all fields etc, just a brief overview).~~ moved to https://github.com/PecanProject/betydb-documentation/issues/12


dlebauer:
#447 provide instructions on configuring the website - perhaps this could be (almost) directly cut and paste to the documentation for the third task in this issue

dlebauer:
image
 Looks great - why not, indeed!

PR #503, opened 2017-04-28
added Reflectance Index as variable type

dlebauer:
Added "Reflectance Index" as option for variable type


gsrohde:
@dlebauer OK if I lower case this type name?

dlebauer:
OK

PR #504, opened 2017-05-02
Add sites cultivars

gsrohde:
Code to add `sites_cultivars` table and related trigger functions.

### DIFF
--- a/db/migrate/20170415183619_add_sites_cultivars.rb
+++ b/db/migrate/20170415183619_add_sites_cultivars.rb
@@ -1,0 +1,81 @@
+class AddSitesCultivars < ActiveRecord::Migration
+  def up
+
+    execute %q{
+
+CREATE TABLE sites_cultivars (
+    id serial8 NOT NULL,
+    site_id bigint NOT NULL CONSTRAINT site_exists REFERENCES sites ON UPDATE CASCADE,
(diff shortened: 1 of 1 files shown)


gsrohde:
@robkooper, @dlebauer Here's a migration to add `sites_cultivars` and related trigger functions.  Detailed explanations are here:

[sites_cultivars.pdf](https://github.com/PecanProject/bety/files/972071/sites_cultivars.pdf)

gsrohde:
@dlebauer @robkooper Please hold off on reviewing the code here—I'm still doing some testing to see if this is working correctly with the BETYdb Rails app.  (You could still look at the PDF document if you want a good understanding of how this is _supposed_ to work.)

dlebauer:
I looked at the doc and got stuck on the set notation. Do you have a
reference for that?

gsrohde:
Sorry—I got carried away with the mathematical logic notation.  (After years of working with this notation, I think more clearly when I write things this way, but perhaps I should write it out in plain English in documents such as this.)

  Basically, the epsilon symbol means "contained in" (or just "in"), the upside-down A means "for all" and the backwards E means "there exists".  So the first line basically says "for all rows `r` in the `sites_cultivars` table, there exists a row `s` in the `sites` table such that `s.site_id = r.site_id`.  (This is the basic foreign-key condition that ensures `sites_cultivars` doesn't contain dangling references to the `sites` table.)  Here's a reference: https://en.wikipedia.org/wiki/List_of_logic_symbols.

PR #514, opened 2017-05-17
Implemented a "contains" parameter for the API for filtering sites.

gsrohde:
@max-zilla You can test out this implementation if you have your own instance of BETYdb you can pull to.  Otherwise, I'll put it up on one of the test sites.

Try, for example,
```
http://localhost:3000/api/beta/sites?containing=42.7941565,-76.11638
```
if you are using the WEBrick Rails server and have the sites data from `ebi_production`.  Note that the numbers are in the order lat,lon as is conventional.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -70,6 +70,15 @@ class Api::BaseController < ActionController::Base
       end
     end
 
+    if (model == Site) && (params.has_key? "containing")
+      lat_lon_text = params['containing']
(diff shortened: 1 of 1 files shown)


max-zilla:
@gsrohde wow, awesome. will try to test tomorrow - this will be great, thanks!

max-zilla:
You can merge this as far as I'm concerned. Once merged we can add a function to the API for extractors to use.

gsrohde:
@max-zilla This change has been deployed.

PR #515, opened 2017-05-18
Updated citation in defaults.yml

dlebauer:


dlebauer:
Corrected the name order. The citation describes betydb.org and it's data so it is appropriate as default.

dlebauer:
This is correct - this citation should be the default, and it should appear on betydb.org. It is up to other users to override this but is appropriate for the PEcAn network

gsrohde:
I had to correct the YAML to quote a string value containing a colon (:).  Otherwise the parser tries to see this a separating a key-value pair.

Issue #521, opened 2017-07-11
Two bugs in sites_cultivars


dlebauer:
* [x] 1. need uniqueness constraint to prevent duplicates of `site_id` ~~(site_id, cultivar_id)~~
* [x] 2. indexing should use the assigned id range for the database. currently appears to start from 1 on all databases.

dlebauer:
To remove duplicates from sites_cultivars

```sql
DELETE FROM sites_cultivars 
WHERE id IN (SELECT id
              FROM (SELECT id,
                             ROW_NUMBER() OVER (partition BY site_id, cultivar_id ORDER BY id) AS rnum
                     FROM sites_cultivars ) t
              WHERE t.rnum > 1);
```

gsrohde:
@dlebauer The duplicate-removal code looks fine.  The bug fixes are in pull request #522, which I've asked @robkooper to review.

gsrohde:
Fixed by release 4.18.

Issue #501, opened 2017-04-08
Travis is failing with 'relation "experiments" does not exist'.


gsrohde:
* [x] Update `test_structure.sql` to match `production_structure.sql`.
* [ ] Put a note in the documentation for upgrading the canonical (EBI) BETYdb Rails site that both `production_structure.sql` and `test_structure.sql` need to be updated whenever the database schema changes.

@robkooper Can you look at this when you get a chance.  For now, I made a branch called `travis_kludge` which is like `master` with the experiments controllers removed.  This makes Travis pass.

robkooper:
Can you link to the failing build?

gsrohde:
https://travis-ci.org/PecanProject/bety/builds/219465407

This is the first build on master that fails for this reason, I believe.  The problem is that the file app/controllers/api/beta/experiments_controller.rb requires the experiments table to be there: On my own copy of BETYdb, if I remove the experiments table, RSpec will crash.  If I then remove this file, RSpec will run again.

robkooper:
Looks like that `bundle exec rake db:structure:load` does not create the experiments table?

gsrohde:
Updated test_structure.sql.  Can be closed after next release.

PR #522, opened 2017-07-13
Fixed id sequence values for sites_cultivars table (issue #521).

gsrohde:
@robkooper Please check that I got the sequence setting code right.

### DIFF
--- a/db/migrate/20170712171513_sites_cultivars_uniqueness.rb
+++ b/db/migrate/20170712171513_sites_cultivars_uniqueness.rb
@@ -1,5 +1,19 @@
 class SitesCultivarsUniqueness < ActiveRecord::Migration
-  def change
-    add_index :sites_cultivars, :site_id, unique: true
+  def up
+    this_hostid = Machine.new.hostid
+
+    execute %{
+      SELECT setval('sites_cultivars_id_seq', GREATEST(1, CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
+
+      ALTER TABLE sites_cultivars ADD CONSTRAINT unique_site_id UNIQUE (site_id);
+    }
+  end
+
+  def down
+
+    execute %{
+      ALTER TABLE sites_cultivars DROP CONSTRAINT unique_site_id;
+    }
+
   end
 end


db/migrate/20170712171513_sites_cultivars_uniqueness.rb
```diff
@@ -1,5 +1,19 @@
 class SitesCultivarsUniqueness < ActiveRecord::Migration
-  def change
-    add_index :sites_cultivars, :site_id, unique: true
+  def up
+    this_hostid = Machine.new.hostid
+
+    execute %{
+      SELECT setval('sites_cultivars_id_seq', GREATEST(1, CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
+
+      ALTER TABLE sites_cultivars ADD CONSTRAINT unique_site_id UNIQUE (site_id);
```
dlebauer: I think the constraint should be `unique (site_id, cultivar_id)` to allow for the case where a site or plot is reused. Otherwise it may break on betydb.org

gsrohde: The `sites_cultivars` table is only for sites that uniquely determine a cultivar.  Right now, this table is empty in `ebi_production`, so adding this constraint won't break anything there.

If we were to allow multiple rows with the same `site_id` value but different `cultivar_id` values, then the trigger functions that automatically determine and set the cultivar associated with a trait based on the trait's site would have no way to operate.  The most it would seem that having such a table could give us would be the ability restrict the possible cultivars for a trait based on its site.

We did talk about making site-cultivar associations time-sensitive so that the combination of a site and date (rather than the site alone) would uniquely determine a cultivar, but we decided simply to use different "sites" (perhaps encompassing exactly the same geographical area) if the cultivar planted in a particular spot changes over time.

Again, for traits associated with sites _not_ in the `sites_cultivars` table, there is no restriction on what cultivar the trait is associated with.

dlebauer:
Got it. Thanks for the clarification. It's hard to keep track!

db/migrate/20170712171513_sites_cultivars_uniqueness.rb
```diff
@@ -1,5 +1,19 @@
 class SitesCultivarsUniqueness < ActiveRecord::Migration
-  def change
-    add_index :sites_cultivars, :site_id, unique: true
+  def up
+    this_hostid = Machine.new.hostid
+
+    execute %{
+      SELECT setval('sites_cultivars_id_seq', GREATEST(1, CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
```
robkooper: that looks correct.

robkooper:
@gsrohde can you fix the failing test.

gsrohde:
@robkooper There's an issue for the failing tests—https://github.com/PecanProject/bety/issues/501.  I haven't figured out what the problem is yet, but they've been failing for some time.

PR #527, opened 2017-07-27
Move beta to v1

gsrohde:
This pull request address https://github.com/terraref/computing-pipeline/issues/320.

The beta path should still work.

The APIPIE docs show both sets of paths (v1 and beta).  I'm not sure yet how to get it only to show the v1 paths.


dlebauer:
I trust you but am curious - how does this keep a /beta path that works? Is there some point in the new code that makes /beta redirect to or otherwise become an alias of /v1?

config/routes.rb
```diff
@@ -259,6 +259,22 @@
       resources :traits, only: :create
     end
   end
+
+  # deprecated "beta" path (same as v1)
+  namespace :api, defaults: { format: 'json' } do
+    namespace :v1, path: :beta do
```
gsrohde: @dlebauer It is this section that makes beta an alias for v1.

PR #541, opened 2017-10-05
Revision to setval fixes in migrations.

gsrohde:
This simplifies the corrected setval calls that were part of PR #538.

More importantly, it revises migration 20170118205944_set_experiment_table_id_seq_vals.rb to accommodate the case where experiments were added before this migration was run.  This probably mainly serves a documentary purpose of serving as a model if this issue (forgetting setval calls when adding a new table) comes up in the future.

### DIFF
--- a/db/migrate/20160412030352_add_sitegroups.rb
+++ b/db/migrate/20160412030352_add_sitegroups.rb
@@ -13,7 +13,7 @@ class AddSitegroups < ActiveRecord::Migration
     change_column :sitegroups, :id, :integer, :limit => 8
     # fix the counter for inserting new records
     execute %{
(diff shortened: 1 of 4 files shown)


gsrohde:
@robkooper I don't really understand your point.  Taking the first change in the first file as an example, changing from
```
GREATEST(1, 1 + CAST(1e9 * #{this_hostid}::int AS bigint))
```
to
```
1 + CAST(1e9 * #{this_hostid}::int AS bigint)
```
really makes no difference unless `1 + CAST(1e9 * #{this_hostid}::int AS bigint)` is less than 1, and that can't happen if `this_hostid` is a non-negative integer.  Moreover, except for the migration `20170118205944_set_experiment_table_id_seq_vals.rb` (where I _do_ use `GREATEST`), the `id` column that setval is being used on (if not the whole table itself) is created by the migration, so there will never already be an entry in the database for that id value.

Issue #542, opened 2017-10-09
Provide a convenient method of adding SQL comments


gsrohde:
Currently, updates to the database schema is effected primarily through the use of Rails migrations.  This helps to ensure that different BETYdb database instances conform to the same database schema.

It is, however, desirable to have a less cumbersome method of updating database comments.

dlebauer:
Currently a lot of the documentation is in the schema metadata for the ebi_production (BETYdb.org) database. Ideally this could be propagated / syncd across instances, or stored in a single canonical location that doesn't require a migration for each edit.

gsrohde:
Pull request #543 implements this feature.

Here's how it works.

Edit the file `db/data/database_comments.sql` to add, change, or delete comments.  (Delete a comment by changing it to the empty string.)

The format of this file is roughly
```
table1_name:
  table_comment: comment for table 1
  column1_name: comment for column 1 of table 1
  column2_name: comment for column 2 of table 1
  ...
table2_name:
  table_comment: comment for table 2
  column1_name: comment for column 1 of table 2
  column2_name: comment for column 2 of table 2
  ...
...
```
(The order of the tables and the order of the key-value pairs under each table is immaterial.)

Then run the Rake task
```
rake bety:db:update_comments
```

If you want to make this the canonical set of comments for the database, check the updated version of `db/data/database_comments.yaml` into the repository.  Then, in the next database release, users will be advised to run the Rake task to update their database comments (just as they are advised to run `rake db:migrate` when the database schema needs to be updated).

The Rake task
```
rake bety:db:dump_comments
```
can be used to produce a YAML file from existing comments.  (Run
```
rake -D bety:db:dump_comments
```
for more information on this task and various run options.)

PR #543, opened 2017-10-11
Sql comments rake task

gsrohde:
This implements the feature in issue #542 using YAML files and Rake tasks.

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -1,3 +1,5 @@
+require 'utilities/sql_comments.rb'
+
 class Api::V1::BaseController < Api::BaseController
   include ApiAuthenticationSystem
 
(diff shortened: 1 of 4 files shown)


dlebauer:
Looks great. Just to confirm, does this feature does the foloowing:

* provides a way of exporting comments in the database
* provides a single canonical place for the comments to be stored (in the .yaml file)
* allows users to import the comments into the local database with a rake task
* provides ruby functions that can be used to query the comments from the database and use this information to populate the API documentation

gsrohde:
It does the first three.  It doesn't provide the Ruby functions to query the comments because they are already there, though it seems from previous comments you have made that they are not going into the API documentation quite where you want them.  I'm considering that a separate issue.

dlebauer:
Approved and Ready to merge, but what is with failing tests?

gsrohde:
The failing tests have something to do with the Travis setup.  I'll make an issue for it if I haven't already.

PR #550, opened 2017-11-02
More docker cleanup

robkooper:
- Pass git data as environment variables so no more need to copy .git folder
- Cleanup of Dockerfile and entrypoint.

hooks/build is what is executed by dockerhub to build the image.


dlebauer:
Looks Awesome!

PR #556, opened 2017-12-15
Specify methods in bulk upload

gsrohde:
Please test for usability (and that it works!).  If you don't want to deploy this on your own, you can test it out at http://pecandev.igb.illinois.edu/jk/ (you can log in as carya).


gsrohde:
Superseded by PR #557.

PR #557, opened 2017-12-15
Bulk upload time of day support + support for specifying methods for each trait.

gsrohde:
This supersedes PR #556, including it and time-of-day support.


gsrohde:
Please test for usability (and that it works!). If you don't want to deploy this on your own, you can test it out at http://pecandev.igb.illinois.edu/jk/ (you can log in as carya).

The two main things to test are:

* Specifying a time with the date for values in the "date" (or "local_datetime, which is now the preferred column name) column, e.g. "2001-04-06T12:34:56" or "2001-04-06 12:34:56".

* Specifying a method for each trait on the "Specify Upload Options and Global Values" page of the Bulk Upload Wizard.

Issue #481, opened 2017-02-06
Issue when creating new formats: Some variable information doesn't stick


serbinsh:
I am noticing that when creating formats via BETYdb web often when defining a variable if you don't put in Name, Unit, Storage Type in at the right time, the information doesn’t stick upon add or update. If you then try to add back and “update” it disappears after sync? Only way around is to delete the variable and re-add but then things get out of order (which isn’t critical)….Just wondered if anyone else found this issue as well?  I know that @istfer has seen the same 

Sometimes show up as yellow when attempting to re-enter the data. (see https://files.gitter.im/PecanProject/bety/Vi8G/Screen-Shot-2017-02-06-at-8.18.21-AM.png)

serbinsh:
<img width="631" alt="screen-shot-2017-02-06-at-8 18 21-am" src="https://cloud.githubusercontent.com/assets/3067369/22666757/f020e7f0-ec87-11e6-8b00-5f10686cec63.png">

gsrohde:
@serbinsh Where is this page?  Give a sample URL for accessing this form.

serbinsh:
@gsrohde this is a run/formats table issue when you are editing associated variables.  For example: https://www.betydb.org/formats/2000000003/edit#

gsrohde:
@serbinsh A change I made almost three years ago broke the editing function.  During this time, the "Save" button essentially did nothing.  (Unfortunately, there is no immediate user feedback that shows whether changes to a row have been saved.  Reloading the page or leaving it and returning to it are two ways to tell whether the information one has entered has been saved, but this is not very convenient.)

Note that you can only update one row at a time.  If you enter information in the "Name", "Unit", "Storage Type", "Column Number" of some row and then, before saving it, you press the "Save" or delete buttons for some other row (or Add a new row), the information you entered in the first row will disappear.

I have at least made the page marginally more usable by preventing the rows from re-sorting when a row is updated.

The yellow highlighting you are seeing is probably some artifact of the browser you are using having to do with its auto-fill function.

PR #562, opened 2018-01-02
Fixes for GH issue #481:

gsrohde:
Fixes route for path "formats/edit_formats_variables" so that responds to "posts"
rather than "gets".  This fixes that basic problem that `formats_variables` relations were not editable—one had to delete and re-add them to update the Name, Unit, Storage Type, or Column Number.

Change format variable deletion link so that it uses the post method instead of
the get method and changed corresponding route accordingly.  (POST is preferred to GET for requests that change data.)

In general, simply routing rules to do with formats and eliminate extraneous
formats_variables resource rule.

Sort format variables by creation time so that they stay in the same place when
edited.  This fixes a usability problem mentioned in issue #481.


gsrohde:
Merged.

Issue #552, opened 2017-11-06
Travis fails trying to create postgis extension.


gsrohde:
The error message is `The command "psql -U bety -d test -c "CREATE EXTENSION postgis;"" failed and exited with 1 during .`

robkooper:
updated to postgres 9.5

PR #568, opened 2018-01-29
Upgrade to rails 4.1

gsrohde:
To deploy this, you will need to:
* Create `config/database.yml` (as usual).
* Copy `public/javascripts/cache/all.js-sample` to `public/javascripts/cache/all.js`.
* Define the environment variable `SECRET_KEY_BASE`.  (For running under Apache http server, I did this by adding a directive of the form
```
    SetEnv SECRET_KEY_BASE [some hex string]
```
inside the `Directory` block for the app.  You can generate a secret key by running `rake secret`.  Alternatively, for testing purposes, you can just edit `config/secrets.yml` and replace `<%= ENV["SECRET_KEY_BASE"] %>` with some hard-coded value.)


gsrohde:
Testers should:
* Test any anomalous behavior against the latest released version of BETYdb
* Make a note _here_ if the anomalous behavior shows up in the upgrade version but not in the latest released version.
* Make a new issue (if no issue already exists) if the anomalous behavior also shows up in the latest released version.

gsrohde:
@dlebauer The RSpec tests pass on my server.  This build problem is connected with issue #552.  I've assigned this to @robkooper.

robkooper:
see #570

dlebauer:
@robkooper this is assigned to you - can you deploy this so it can be tested?

robkooper:
took an exisiting instance of pecan, switched to this branch, ran`bundle install` set the secret key and now I end up with an error, no log files. Not sure what to do next, I can show it after the TERRA meeting.

max-zilla:
evaluation link: http://pecan-dev.ncsa.illinois.edu/bety/

gsrohde:
@robkooper I the 500 error message when I try to visit the evaluation site.  Is there a place I can ssh into to view the Rails logs?

dlebauer:
I checked out the website - went through the process of creating new citations, sites, treatments, trait. Everything works as expected -

robkooper:
@max-zilla let us know when you checked the rest api used in terra project.

max-zilla:
I tested the endpoints used in terrautils (sites, experiments, traits, cultivars) and everything looked good with no errors.

gsrohde:
Found one error—the model "show" pages, e.g. http://pecan-dev.ncsa.illinois.edu/bety/models/3.  Will fix before pulling.

gsrohde:
Also, show pages for inputs and posteriors, e.g. http://pecan-dev.ncsa.illinois.edu/bety/inputs/4 and http://bety-dev:3000/posteriors/523.  These are all the same problem, though, which is easily fixed.

In all cases, the problem is caused by an obsolete function API for limiting the number of associated dbfiles shown to 100.  I can either (1) replace the function call with one that still works in Rails 4.1; (2) eliminate the limit (in the EBI database, no instance of any of these types ever has more the _3_ associated dbfiles!); (3) keep the limit (as in (1)), but add a warning that not all associated files are shown in the case that there are ever more than 100 of them.

gsrohde:
I went with option 2 (eliminating the 100-item limit) to fix the problem mentioned above.

PR #569, opened 2018-01-29
added sites_cultivars.rb

dlebauer:
I would like to have sites_cultivars associations be avaialable via api when associations_mode =full_info. I am taking a first pass at enabling this


betydb:
I would like to have sites_cultivars associations be avaialable via api
when associations_mode =full_info. I am taking a first pass at enabling this

dlebauer:
If I recall, the minimum functionality that we needed here was to be able to access the sites_cultivars relationships. Currently that is difficult to do. This would be nice to have if it is ready to deploy but at this point is low priority.

PR #572, opened 2018-02-07
unicorn and bety root path

robkooper:
Added gem + configuration to run bety using unicorn. This is a frontend to
multiple instances of bety running, which will speed up the response time.

Added ability to set the path to the bety application using an environment
variable.


robkooper:
To test this you can do the following:

Load the database, this will take about 15 minutes.
```
docker run -t -i --rm --name postgres -v /Users/kooper/docker/postgres:/var/lib/postgresql/data mdillon/postgis:9.5
docker run -t -i --rm --link postgres pecan/bety:unicorn initialize
```

Start the BETY app
```
docker run -t -i --rm --link postgres --port 8000:8000 pecan/bety:unicorn unicorn
```
Now you can connect to http://localhost:8000/

You can also use the option autoserver instead of unicorn. This will automatically migrate the database and start a server using unicorn. This can be used with programs that will monitor docker and automatically update a container with the newest version, for example https://github.com/v2tec/watchtower

```
docker run -t -i --rm --link postgres --port 8000:8000 pecan/bety:unicorn autoserver
```

robkooper:
You can also connect to an external database. First create a database.yml file that contains the information to connect to the remote database. Next start the service using

```
docker run -t -i --rm -v ${PWD}/mydatabase.yml:/home/bety/config/database.yml --publish 8000:8000 pecan/bety:unicorn autoserver
```

max-zilla:
followed your instructions and it generally looked good.

gsrohde:
@dlebauer Given David Slater's citation of serious security flaws in Docker, do we want to keep using it?

robkooper:
@gsrohde do you have a list of these issues? Yes we will want to use docker in many other projects.

dlebauer:
David Slater said:

> we do not support docker. Docker has a serious security flaw where a
regular user can run things as root. This is why we do not install it
anywhere or on the cluster. We are looking at singularity, it is supposed
to not have this problem.

robkooper:
Yes docker runs as root, this however is only an issue if you let people run arbitrarily on a machine, i.e. a hpc cluster, which is exactly where you want to use docker.
Docker does work really well on systems where we have more control over who runs what. In this case you would never want to run BETY inside the HPC cluster, but in a controlled location next to the cluster. This is a perfect use case for docker, where the user does not have to worry about the dependencies.

robkooper:
declining this in favor of work I'm doing to merge this into the bety 5 branch.

PR #573, opened 2018-02-07
MIGRATION: Relax constraints on citations table

ashiklom:
- Change `citations.title` and `citations.journal` columns to unlimited variable width `text` type.
- Remove format constraint on `citations.pg`.


dlebauer:
Is it worth requiring a migration so that we can use letters to number pages?

I'd just assume drop all columns except "Author, year, title, journal, doi"

PR #586, opened 2018-05-15
Add fix database id script, several other database maintenance scripts, and a migration to add missing foreign-key constraints and update existing ones to use the CASCADE update rule.

gsrohde:
The main script in this migration is `fix_database.rb`.  This script gets from the user a set of database connection parameters, an output file name, and the machine id that the database should use.  It then writes out SQL statements to the given file to update the sequence objects for public tables so that new ids are allocated in the range appropriate for the given machine id.  It also writes out statements to update rows with ids in the 99 billion range, resetting the ids to use the appropriate range.

The migration should be run first.  It adds all missing foreign-key constraints.  It also updates several existing constraints to use the "CASCADE" update rule.  This way, when referred-to ids are updated, the referring columns are updated appropriately as well.  (Without this, updating the referred-to ids would be blocked in most cases.)  The choice of which existing foreign-key constraints to update to use the CASCADE rule was guided by the set of existing id problems in the TERRA-MEPP database.  (Other foreign-key constraints can have their "update" rule changed as needed.)

In adding new foreign-key constraints, there was always a question of what _update_ rule and what _delete_ rule to use.  In all cases, I used "CASCADE" as the update rule.  "SET NULL" and "SET DEFAULT" were ruled out on the grounds that information would be lost.  Both "NO ACTION" and "RESTRICT" would prevent information loss by preventing the update if any foreign keys refer to the rows whose ids are being update.  In the past, I haven't usually bothered to set the update rule to anything other the default "NO ACTION" since it seemed unlikely that we would ever need to update the ids of rows.  But now that we have a use-case for this (fixing mis-allocated ids), "CASCADE" seems the appropriate rule—for updates, this never results in the loss of important information.

The _delete_ rules, using "CASCADE" _does_ result in information loss, so in most cases I left the delete rule as the default "NO ACTION", which prevents a referred-to row from being deleted.  The exception was join tables: if a row referred to by a join table row is deleted, it almost always makes sense to delete the join table row as well.  In this case, "CASCADE" is the appropriate delete rule.

The constraint on `dbfiles` to ensure the row referred to by `container_type` and `container_id` is a special case.  It can't be implemented with a foreign-key constraint and must be implemented using a trigger function instead.  The migration updates the trigger function to update the values in the `container_id` column when the referred-to ids are updated.


gsrohde:
@robkooper Given your concerns about adding new foreign-key constraints unless and until we clean up all databases that data may be pulled from, I'd like you for now to just evaluate _this_ plan for fixing problems in the `bety_mepp` database:

1. Take the terra-mepp site off line.  (@justinmcgrath , @dlebauer ) Do I need to announce this?  Can I just do `sudo apachectl stop`?
2. "git fetch" this branch (`add_fix_db_id_script`) and switch to it.  (Alternatively, just manually copy all the files in this pull request—they're all new.)
3. Run the migration included in this PR to ensure that all foreign-key constraints needed for the id-fixing script are in place.
4. (optional) Run `find_missing_constraints.rb` to make sure we didn't miss anything.
5. Run `fix_database.rb` (specifying machine id 8 and the correct database connection parameters) to generate the file of SQL statements that will fix the problems.
6. Run the generated file against the `bety_mepp` database.
7. Revert the migration.
8. Revert the code to the last release.
9. Bring the site back on line.

By the way, the only top-level scripts in `/script/db_maintenance` are `fix_database.rb` and `find_missing_constraints.rb` (and `validate_constraints.rb`, which isn't immediately relevant to the problem at hand).  The dependency tree for `fix_database.rb`, for example, is

```
fix_database.rb
    id_fixer.rb
        enhanced_connection.rb
            postgres_helpers.rb
            connection_information.yml (if given)
```
The other two scripts depend directly on `enhanced_connection.rb`.

db/migrate/20180510184222_update_foreign_key_constraints.rb
```diff
@@ -0,0 +1,311 @@
...
+new. */
+
+ALTER TABLE "posteriors_ensembles" DROP CONSTRAINT "fk_posteriors_ensembles_ensembles_1";
+ALTER TABLE "posteriors_ensembles" ADD CONSTRAINT "fk_posteriors_ensembles_ensembles_1" FOREIGN KEY ("ensemble_id") REFERENCES "ensembles" ("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+ALTER TABLE "runs" DROP CONSTRAINT "fk_runs_ensembles_1";
+ALTER TABLE "runs" ADD CONSTRAINT "fk_runs_ensembles_1" FOREIGN KEY ("ensemble_id") REFERENCES "ensembles" ("id") ON UPDATE CASCADE;
+
+ALTER TABLE "inputs" DROP CONSTRAINT "fk_inputs_inputs_1";
+ALTER TABLE "inputs" ADD CONSTRAINT "fk_inputs_inputs_1" FOREIGN KEY ("parent_id") REFERENCES "inputs" ("id") ON UPDATE CASCADE         not valid;
```
dlebauer: how did you decide which tables to use `not valid` to allow the constraint to be set even on tables where the constraint condition is not met? Is it only based on the tables in bety_mepp that violate the constraints?

db/migrate/20180510184222_update_foreign_key_constraints.rb
```diff
@@ -0,0 +1,311 @@
...
+ALTER TABLE "posteriors_ensembles" ADD CONSTRAINT "fk_posteriors_ensembles_posteriors_1" FOREIGN KEY ("posterior_id") REFERENCES "posteriors" ("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+ALTER TABLE "inputs_runs" DROP CONSTRAINT "fk_inputs_runs_runs_1";
+ALTER TABLE "inputs_runs" ADD CONSTRAINT "fk_inputs_runs_runs_1" FOREIGN KEY ("run_id") REFERENCES "runs" ("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+ALTER TABLE "likelihoods" DROP CONSTRAINT "fk_likelihoods_runs_1";
+ALTER TABLE "likelihoods" ADD CONSTRAINT "fk_likelihoods_runs_1" FOREIGN KEY ("run_id") REFERENCES "runs" ("id") ON UPDATE CASCADE;
+
+ALTER TABLE "citations_sites" DROP CONSTRAINT "fk_citations_sites_sites_1";
+ALTER TABLE "citations_sites" ADD CONSTRAINT "fk_citations_sites_sites_1" FOREIGN KEY ("site_id") REFERENCES "sites" ("id") ON DELETE RESTRICT ON UPDATE CASCADE      not valid;
```
dlebauer: why `on delete restrict` instead of `on delete cascade`?

db/migrate/20180510184222_update_foreign_key_constraints.rb
```diff
@@ -0,0 +1,311 @@
...
+    EXECUTE PROCEDURE forbid_dangling_#{container_type.downcase}_references();
+
+}
+    end
+  end
+
+  def self.down
+	execute %q{
+ALTER TABLE "posteriors_ensembles" DROP CONSTRAINT "fk_posteriors_ensembles_ensembles_1";
+ALTER TABLE "posteriors_ensembles" ADD CONSTRAINT "fk_posteriors_ensembles_ensembles_1" FOREIGN KEY ("ensemble_id") REFERENCES "ensembles" ("id") ON DELETE CASCADE;
```
dlebauer: why not on update cascade?

db/migrate/20180510184222_update_foreign_key_constraints.rb
```diff
@@ -0,0 +1,311 @@
...
+ALTER TABLE "ensembles" DROP CONSTRAINT "fk_ensembles_workflows_1";
+ALTER TABLE "ensembles" ADD CONSTRAINT "fk_ensembles_workflows_1" FOREIGN KEY ("workflow_id") REFERENCES "workflows" ("id");
+
+ALTER TABLE "yields" DROP CONSTRAINT "fk_yields_treatments_1";
+ALTER TABLE "yields" ADD CONSTRAINT "fk_yields_treatments_1" FOREIGN KEY ("treatment_id") REFERENCES "treatments" ("id");
+
+ALTER TABLE "ensembles" DROP CONSTRAINT "fk_ensembles_workflows_1";
+ALTER TABLE "ensembles" ADD CONSTRAINT "fk_ensembles_workflows_1" FOREIGN KEY ("workflow_id") REFERENCES "workflows" ("id");
+
+/* Drop all entirely new constraints. */
```
dlebauer: why are you dropping the constraints here?

dlebauer:
@gsrohde you don't need to announce bringing the terra mepp instance offline. 

I don't really understand the order of operations for the network of PEcAn databases and defer to Rob on that. However, I thought that the whole point of the 'not valid' clause was that you could add constraints to the tables even if existing rows are in violation of the constraint. Would using the `not valid` clause on every constraint resolve issues with syncing? In any case, my vote is for as many constraints as possible.

db/migrate/20180510184222_update_foreign_key_constraints.rb
gsrohde: Simply because that's how things are now—the down migration is meant to restore the database structure to its original state, that is, the state it was in before the migration was applied.

db/migrate/20180510184222_update_foreign_key_constraints.rb
gsrohde: Yes.  Since the motivation for this migration was fixing `bety_mepp`, I tried it out on a copy of `bety_mepp` and applied `NOT VALID` as needed.  (There may have also been cases where the constraint I was adding `ON UPDATE CASCADE` to was already marked `NOT VALID` and I didn't see fit to change it.)

_Most_ of the problems in `bety_mepp` arise when a table is imported from another database (usually `ebi_production` on `ebi-forecast`) but the tables it refers to are not.  Foreign-key constraints are lifted during imports and they aren't enforced for the imported data (though they are put back in place for "new" data).  On the other hand, a uniqueness constraint violation can cause a particular table not to get imported even if other tables are.

There are a few cases where the constraints would be violated even if all data were successfully imported from all machines.

gsrohde:
@robkooper I added a README file to the `scripts/db_maintenance` directory and added extensive help information to each of the scripts (available using the `--help` and `--man` options).  Let me know if you have enough here to go by.

PR #590, opened 2018-06-10
recaptcha update

### LINKED ISSUE #588: reCaptcha VI Shutdown
Currently new users cannot sign up for a new account on bety because recaptcha v1 shutdown march 31st and needs to be updated. 

screen shot 2018-06-04 at 10 49 11 am

https://developers.google.com/recaptcha/docs/faq#what-happens-to-recaptcha-v1

robkooper:
Update to latest version of recaptcha.

future should remove hard coded recaptcha keys and use environment variables.

fixes #588


app/controllers/users_controller.rb
```diff
@@ -46,7 +46,7 @@ def create
     if Rails.env == "test"
       success = @user && @user.save
     else
-      success = verify_recaptcha(:model => @user, :message => "Please re-enter the words from the image again.") && @user && @user.save
+      success = verify_recaptcha(:model => @user) && @user && @user.save
     end
```
gsrohde: Is :message redundant or no longer necessary?

app/views/users/new.html.erb
```diff
@@ -76,7 +76,7 @@
         <em>* Any request for page access higher then Viewer or data access higher then External researcher must be approved by an administrator</em>
 
         <% if Rails.env != "test" %>
-          <%= recaptcha_tags ssl: true %>
+          <%= recaptcha_tags %>
         <% end %>
```
gsrohde: Is the ssl key not needed?

robkooper: This is no longer an option see https://github.com/ambethia/recaptcha#recaptcha_tags

app/controllers/users_controller.rb
robkooper: it complained during runtime, so I removed, however it should be possible see https://github.com/ambethia/recaptcha#verify_recaptcha

PR #599, opened 2018-09-28
fix download url and oter small docker fixes

robkooper:
URL used to download initial version of BETY was wrong (used pecanall instead of pecan/all).

make sure to use latest in docker-compose not special tag.


robkooper:
yes, it auto builds the following:
branch : master -> latest
branch : develop -> develop
tag : /^betydb_(.+)/ -> \1 (specific version)

Issue #603, opened 2018-10-11
Method completion doesnt seem to work in web interface


serbinsh:
<img width="1040" alt="screen shot 2018-10-11 at 10 37 30 am" src="https://user-images.githubusercontent.com/3067369/46812957-cf58b400-cd43-11e8-9897-c59254749bc8.png">

I am trying to select the method for these entries.  It says I need to select from a list of options but no list appears?  Is there a way to set via the file?  I can't find the appropriate column name in the documentation?

serbinsh:
For this upload I just did

```
UPDATE traits SET method_id=2000000002 WHERE traits.citation_id=2000000027;
```

serbinsh:
@dlebauer what do you mean? The methods table has various methods in it and I have been trying to use it more to differentiate traits that have the same name but are measured / estimated with different approaches, etc

dlebauer:
I just meant were there any rows in your methods table, which it sounds like there are. Your use is in line with how it is intended.

gsrohde:
@serbinsh It seems that I broke this in release 5.0—it worked in release 4.21.  It was a one-line fix which is currently in pull request #604, and which I'll release as soon I can get to it after it is approved.  By the way, the Bulk Upload Wizard does not support a `method` column.

gsrohde:
@serbinsh I've released a fix for this (BETYdb 5.0.5).

PR #604, opened 2018-10-22
Fix for issue #603.

gsrohde:


gsrohde:
@robkooper It occurred to me I should have added to the change log.  I've done so.  OK?

robkooper:
good catch, feel free to release.

Issue #605, opened 2018-11-08
Allow API access to public data without a key 


dlebauer:
Currently, anyone can access public data with a 'public' key, i.e.

The guest user API key is 9999999999999999999999999999999999999999, and this can be used by anyone to access the metadata tables as well as to any traits records with access_level = 4, i.e.:

https://betydb.org/api/v1/treatments?key=9999999999999999999999999999999999999999

It is difficult to remember this key, so one option would be to change the key to something like 'publickey', but @robkooper suggested the following feature request:

Change the API so that the `key` parameter is optional; if no key is provided, the request is handled as the guest user.

serbinsh:
@dlebauer how do I implement the "guest user" on modex bety so we can enable that API key to access the DB?  I tried it out on our system and received an API warning 

```
{
  "metadata": {
    "URI": "https://modex.bnl.gov/bety/api/beta/species?limit=250&scientificname=Citrus+%C3%97+sinensis",
    "timestamp": "2019-01-15T08:22:41.820-05:00",
    "count": null
  },
  "errors": "authentication failed: For key-less access to the API, you must set up the guest user account."
}
```

robkooper:
Make sure guest user does not exist (all commands are run on bety database):
```
SELECT count(id) FROM users WHERE login='guestuser';
```
If this is non zero you have a guest user already

Add guest user:
```
INSERT INTO users (login, name, email, crypted_password, salt, city, state_prov, postal_code, country, area, access_level, page_access_level, created_at, updated_at, apikey, remember_token, remember_token_expires_at) VALUES ('guestuser', 'guestuser', 'betydb+${ID}@gmail.com', '994363a949b6486fc7ea54bf40335127f5413318', 'bety', 'Urbana', 'IL', '61801', 'USA', '', 4, 4, NOW(), NOW(), NULL, NULL, NULL);
```

serbinsh:
Thanks @robkooper Here is my current issue

```
[sserbin@modex ~]$ psql -d bety -U bety
psql (9.4.20)
Type "help" for help.

bety=> SELECT count(id) FROM users WHERE login='guestuser';
 count
-------
     0
(1 row)

```

```
bety=> INSERT INTO users (login, name, email, crypted_password, salt, city, state_prov, postal_code, country, area, access_level, page_access_level, created_at, updated_at, apikey, remember_token, remember_token_expires_at)
[...]

PR #606, opened 2018-11-26
Allow keyless guests

gsrohde:
Implements feature requested in Issue #605.

### DIFF
--- a/app/services/api_authentication_system.rb
+++ b/app/services/api_authentication_system.rb
@@ -3,7 +3,11 @@ module ApiAuthenticationSystem
 
   # Override default access_denied action.
   def access_denied
-    @errors = "authentication failed"
+    if @errors
+      @errors = "authentication failed: " + @errors
+    else
+      @errors = "authentication failed"
+    end
     render status: 401
   end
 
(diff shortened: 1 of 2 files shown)


app/services/api_authentication_system.rb
```diff
@@ -19,4 +23,21 @@ def permissions(action_name, resource)
...
   end
 
+  # Override "login_from_api_key" so that if no key is given or the given key is
+  # invalid, the user is logged in as the guest user.
+  def login_from_api_key
+    key = params[:key]
+    if key.nil?
+      u = User.find_by_login('guestuser')
+    else
+      u = User.find_by_apikey(params[:key]) || User.find_by_login('guestuser')
```
robkooper: If a key is given and it is wrong, I think it should return an error that states it is an invalid login. Otherwise the user might be confused why their login does not allow them to do certain actions in case the API is wrong (since it becomes guestuser).

gsrohde: Better?

app/services/api_authentication_system.rb
```diff
@@ -28,13 +29,15 @@ def permissions(action_name, resource)
...
+      if u.nil?
+        @errors = "For key-less access to the API, you must set up the guest user account."
+      end
     else
-      u = User.find_by_apikey(params[:key]) || User.find_by_login('guestuser')
-    end
-
-    if u.nil?
-      @errors = "You must either use a valid API key or set up the guest user account."
+      u = User.find_by_apikey(params[:key])
```
robkooper: nitpicking but can use `key` variable here to prevent an extra lookup.

gsrohde: No problem--that was my intention but then I forgot to; and your nitpicking led me to notice and fix a more serious problem.

Issue #597, opened 2018-09-25
`load.bety.sh` does not automatically track PEcAn updates in Docker container


ashiklom:
Because `load.bety.sh` is pulled in via a `curl` command in the [Dockerfile](https://github.com/PecanProject/bety/blob/master/Dockerfile#L33-L36), revisions to `load.bety.sh` will _not_ trigger updates to the `bety` Docker container. This means it's very easy for the `bety` Docker container to become out of date with changes to its initialization procedure.

The most straightforward way to fix this is to move the `load.bety.sh` script from `pecan` into `bety` (which, to me, just generally makes more sense).

robkooper:
agree, we can move load/dump scripts to bety instead of pecan.

Issue #505, opened 2017-05-03
New generic table for misc. information polymorphic relations 


dlebauer:
Table name: ~~miscelaneyiis~~ please propose ... 

Use cases:

* adding synonyms to variables
* adding additional metadata to cultivars (e.g. lineage, etc)
* PEcAn, add all parameters that we currently store in xml
* generally whereever we have notes that we use to store additional info
* (eventually) could deprecate many fields that are sparsely populated

### Implementation (Option 3)

**Fields**

* type: table name
* type_id: id in table name
* value: json

unique type, type_id

**Plus standard ruby stuff**

* id: bigint
* updated_at
* created_at

---

### Other options originally considered

**Option 1**
Four fields:

* type: table name  (key unique)
* type_id: id in table name
* key: character
* value: charcater

unique type, type_id, key

**Option 2**

* type: table name   
* type_id: id in table name
* key: character
* value_string:
* value_float:
* value_int:
* ...

unique type, type_id, key

**Option 2.1**

* type: table name   
* type_id: id in table name
* key: character
* type: (char) data type
* value: (char)

**Option 3**

* type: table name
* type_id: id in table name
* value: json

unique type, type_id

**Plus standard ruby stuff**

* id: bigint
* updated_at
* created_at

dlebauer:
From meeting today, it sounds like Option 3 is preferred.

Here is a description of the json data type https://www.postgresql.org/docs/9.3/static/datatype-json.html and a list of json operations: https://www.postgresql.org/docs/9.3/static/functions-json.html in the postgres documentation.

robkooper:
New database will be called attributes

Make sure to use something following like this to upsert into the database:
```
INSERT INTO attributes(type, type_id, value) VALUES ('sites', '1', '{"set":{"some_prop": 88}}')
ON CONFLICT (rec_id) DO UPDATE jsonb_set(attributes.value, array['set','some_prop'], '88', true) WHERE type='sites' AND type_id=1;
```

robkooper:
Table will be called attributes

ashiklom:
A possible application for loading raster formats, per request of @serbinsh:

- Create a new `format` called `raster_projection_x`
- Describe the format in `attributes` table: `INSERT INTO attributes (type, id, value) VALUES ("formats", XXXX, "{'gis' : {'epsg' : 12345, 'reference_lat' : 45, 'reference_lon' : -83}, 'comment' : 'I hate this format'''}");`
[...]

PR #609, opened 2018-11-30
attributes migration

robkooper:
Add a new table that allows for attributes in json form #505
Add dump/load bety scripts from PEcAn to BETY #597

### DIFF
--- a/db/migrate/20181129000515_create_attributes_table.rb
+++ b/db/migrate/20181129000515_create_attributes_table.rb
@@ -15,7 +15,7 @@ class CreateAttributesTable < ActiveRecord::Migration[5.1]
     reversible do |dir|
       dir.up do
         execute %{
-          SELECT setval('attributes_id_seq', GREATEST(1, 1 + CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
+          SELECT setval('attributes_id_seq', 1 + CAST(1e9 * #{this_hostid}::int AS bigint), FALSE);
           ALTER TABLE "attributes"
             ALTER COLUMN created_at SET DEFAULT utc_now(),
             ALTER COLUMN updated_at SET DEFAULT utc_now(),
--- a/db/migrate/20181129000515_create_attributes_table.rb
+++ b/db/migrate/20181129000515_create_attributes_table.rb
@@ -1,0 +1,27 @@
+class CreateAttributesTable < ActiveRecord::Migration[5.1]
+  def change
+    this_hostid = Machine.new.hostid
+
+    create_table :attributes, id: :bigint do |t|
+      t.string :container_type, null: false
+      t.integer :container_id, limit: 8, null: false
+      t.jsonb :value, null: false, default: '{}'
+      t.timestamps
+    end
+
+    add_index :attributes, :container_id
+    add_index :attributes, :value, using: :gin
+
+    reversible do |dir|
+      dir.up do
+        execute %{
+          SELECT setval('attributes_id_seq', GREATEST(1, 1 + CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
+          ALTER TABLE "attributes"
+            ALTER COLUMN created_at SET DEFAULT utc_now(),
+            ALTER COLUMN updated_at SET DEFAULT utc_now(),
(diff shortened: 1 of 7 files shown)


db/migrate/20181129000515_create_attributes_table.rb
```diff
@@ -0,0 +1,27 @@
+class CreateAttributesTable < ActiveRecord::Migration[5.1]
+  def change
+    this_hostid = Machine.new.hostid
+
+    create_table :attributes, id: :bigint do |t|
+      t.string :container_type, null: false
+      t.integer :container_id, limit: 8, null: false
+      t.jsonb :value, null: false, default: '{}'
```
gsrohde: Would it make sense to constrain `value` in some way?  I assume this is just supposed to be a collection of key-value pairs.

db/migrate/20181129000515_create_attributes_table.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.timestamps
+    end
+
+    add_index :attributes, :container_id
+    add_index :attributes, :value, using: :gin
+
+    reversible do |dir|
+      dir.up do
+        execute %{
+          SELECT setval('attributes_id_seq', GREATEST(1, 1 + CAST(1e9 * #{this_hostid}::int AS bigint)), FALSE);
```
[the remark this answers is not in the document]
robkooper: done

db/migrate/20181129000515_create_attributes_table.rb
gsrohde: Yes, but really _any_ JSON document?

robkooper: yup, i don't want to limit this table. The whole idea of this table is for people to add any data to any existing row in the database.

gsrohde: So even though we call this table "attributes", the json really could be anything, it doesn't even have to be a hash, yes?

robkooper: it has to be json format that is all we know. We don't know what it will contain.

gsrohde: OK, I'll just assume you know what you're doing.

robkooper:
Waiting on merging this issue to see if we want to rush this before AGU

gsrohde:
@dlebauer Rob addressed one minor coding nit that I flagged.  Regarding the larger concern having to do with lack of constraints on the `value` column, I don't think I really am in a position to weigh in on this question.  I'm not versed in the theory or philosophy of using relational databases in this way, and I wasn't in on any discussion of what the motive was for adding this table and how it would be used.  From what I could glean from the simple fact that this is called an _attributes_ table, I take it that this was seen as a method for adding "sparse" attributes to various database relations in an ad hoc way.  If this is the case, I would think there would at least be the requirement that the value be a JSON object (as opposed to an array or a primitive) with the keys serving as the attribute names and perhaps no restriction on the values.

I'm also guessing that this new table was motivated by needs coming from PEcAn and that the BETYdb web app will simply provide the migration and will otherwise ignore these changes.  Under this assumption, I'm going to go ahead and click Approve changes since I think the changes are largely out of the scope of both my expertise and what I am in charge of.

gsrohde:
@robkooper @dlebauer Actually, I do have one concern not related to any previously expressed concerns—namely, that this requires at least PostgreSQL 9.4 and the official documentation only calls for version 9.3 or did until very recently.  (I somehow can access the technical documentation link, so I can't check what it says now.)  I think all the deployments I work with—EBI, bety-mepp, and terra-ref—now are using 9.4 or later, but I don't know if this is the case with all the other public instances.

robkooper:
https://github.com/PecanProject/betydb-documentation/blob/master/installing_betydb/installing-betydb-web-application.md talks about PostgreSQL 9.3 or greater. other places in the documentation talks about postgresql 9.4

gsrohde:
@robkooper Yes, I noticed that, when the site eventually came up for me.  I also made this change on the `bookdown` branch of the documentation, which will eventually replace the master branch when we deploy the bookdown version to some official place.  (It can't replace it now, because GitBook will choke on some of the bookdown syntax.)  If you think it make sense to, I'll rename the `bookdown` branch to `develop`.

PR #620, opened 2019-01-11
Fix to make sure dbfiles have valid filename

robkooper:
A fix for a critical vulnerability reported by Danny Rosseau at Carve Systems (www.carvesystems.com).

### DIFF
--- a/app/models/db_file.rb
+++ b/app/models/db_file.rb
@@ -32,6 +32,11 @@ class DBFile < ActiveRecord::Base
     self[:created_user_id] = user_id
     self[:updated_user_id] = user_id
     if upload # Uploaded file
+      # check to see it is a valid filename
+      if upload.original_filename.match(/^[a-zA-Z0-9._\-]+$/).nil?
+        raise("Invalid filename \"#{upload.original_filename}\"")
+      end
+
       self[:file_name] =  upload.original_filename
       self[:md5] = Digest::MD5.file(upload.path).hexdigest
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,11 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 
 ### Fixes
 
-- A fix for a critical vulnerability reported by Danny Rosseau at Carve Systems (www.carvesystems.com).
+- #611 : could not close alerts
+- #585 : Pressing "Show" on the Covariates Edit page attempts an update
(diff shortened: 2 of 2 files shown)


robkooper:
@gsrohde also update CHANGELOG to capture older commits/pull requests.

app/models/db_file.rb
```diff
@@ -32,6 +32,11 @@ def setup(user_id, upload, args = nil)
     self[:created_user_id] = user_id
     self[:updated_user_id] = user_id
     if upload # Uploaded file
+      # check to see it is a valid filename
+      if upload.original_filename.match(/^[a-zA-Z0-9._\-]+$/).nil?
```
dlebauer: filenames can only contain letters, numbers, periods and dashes (-)?

app/models/db_file.rb
```diff
@@ -32,6 +32,11 @@ def setup(user_id, upload, args = nil)
...
+        raise("Invalid filename \"#{upload.original_filename}\"")
```
dlebauer:
```suggestion
        raise("Invalid filename \"#{upload.original_filename}\"; filenames can only contain letters, numbers, periods and dashes.")
```

app/models/db_file.rb
robkooper: correct

gsrohde: The backslash is extraneous.  But harmless—basically a no-op (I think).

app/models/db_file.rb
```diff
@@ -32,6 +32,11 @@ def setup(user_id, upload, args = nil)
     self[:created_user_id] = user_id
     self[:updated_user_id] = user_id
     if upload # Uploaded file
+      # check to see it is a valid filename
+      if upload.original_filename.match(/^[a-zA-Z0-9._\-]+$/).nil?
+        raise("Invalid filename \"#{upload.original_filename}\"; filenames can only contain letters, numbers, periods and dashes.")
```
gsrohde: You left out underscores.

gsrohde:
@robkooper This (with the minimal changes I suggest) looks fine to me.  I'm curious, though, to understand the "critical vulnerability" this addresses, information about which I couldn't find on the cited site.

app/models/db_file.rb
dlebauer:
```suggestion
        raise("Invalid filename \"#{upload.original_filename}\"; filenames can only contain letters, numbers, periods, underscores, and dashes.")
```

robkooper:
I have the information about the issue and can share it with you when you get to the office. Until this is released and people have it deployed I'd like to not make it to clear.

Issue #618, opened 2019-01-10
Error page after trying to download data from search page in web UI


dlebauer:
When I go to betydb.org (or do similar w/ terraref.ncsa.illinois.edu/bety) and I 

1. enter "panicum sla" into the search box and hit return
   - the results show up as expected
2. on the results page, when I click 'download results'
   - I expect a CSV of the results to be downloaded
   - instead, I get an error:

![image](https://user-images.githubusercontent.com/464871/51002487-a4bbfa00-14f0-11e9-88e6-1afd4a64230c.png)

Is this a known bug? Related to rails updates?

PR #622, opened 2019-01-11
Fix for issue 618.

gsrohde:
One-line fix for issue #618.


robkooper:
Missing changelog, will add to big changelog update in #620

Issue #621, opened 2019-01-11
security issue when uploading files


robkooper:
There is an issue when uploading files.

A fix for a critical vulnerability reported by Danny Rosseau at Carve Systems (www.carvesystems.com).

Issue #611, opened 2018-12-03
Close button in alerts no longer working.


gsrohde:
For example, the green alert that appears when you log in successfully can't be closed by clicking the X.  This problem cropped up when the Rails asset pipeline was enabled and is easily fixed.

Issue #596, opened 2018-08-29
Crop Model Maps links are broken


gsrohde:
The links to the crop model maps are broken on the EBI site.  They work on the Terra-MEPP site.

dlebauer:
fixed by #607

Issue #585, opened 2018-05-14
Pressing "Show" on the Covariates Edit page attempts an update.


gsrohde:

gsrohde:
Fixed in release BETYdb 5.1.0.  Closing.

PR #630, opened 2019-01-14
release 5.1.0

robkooper:
### Fixes

- #611 : could not close alerts
- #585 : Pressing "Show" on the Covariates Edit page attempts an update
- #621 : A fix for a critical vulnerability reported by Danny Rosseau at Carve Systems (www.carvesystems.com).
- #596 : Crop Model Maps links are broken
- #605 : Allow for keyless API access (using guestuser access)
- #618 : Can not download results of search as csv

### DIFF
--- a/app/models/bulk_upload_data_set.rb
+++ b/app/models/bulk_upload_data_set.rb
@@ -1242,8 +1242,12 @@ class BulkUploadDataSet
           t.site_id = row['site_id']
 
           # Modify each Trait class instance so that date strings are
-          # interpreted as being in the time zone of the trait site
-          # (or UTC, if the trait site time_zone column is null)
+          # interpreted as being in the time zone of the trait site (or UTC, if
+          # the trait site time_zone column is null).  Note that in the Bulk
+          # Upload Wizard, the validation step will prevent reaching this code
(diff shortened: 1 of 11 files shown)


app/models/bulk_upload_data_set.rb
```diff
@@ -1242,8 +1242,12 @@ def insert_data
...
 
           # Modify each Trait class instance so that date strings are
-          # interpreted as being in the time zone of the trait site
-          # (or UTC, if the trait site time_zone column is null)
+          # interpreted as being in the time zone of the trait site (or UTC, if
+          # the trait site time_zone column is null).  Note that in the Bulk
+          # Upload Wizard, the validation step will prevent reaching this code
+          # if a site without time zone is in the data file.  But nothing
+          # prevents a site without time zone from being specified
+          # interactively.
```
dlebauer: @gsrohde you mean time zones aren't assigned automatically based on location?

gsrohde: @dlebauer That is correct, unfortunately.  When I added the `time_zone` column, I did create and use table of time-zone/location data to facilitate automatically populating the `time_zone` column for existing sites, but there is nothing built-in to do this when adding a new site.

Issue #551, opened 2017-11-06
JavaScript errors on every page change


gsrohde:
Every time the user goes to a new page, to errors recur:

    Uncaught ReferenceError: GUnload is not defined at onunload

and

    Uncaught Error: jquery-ujs has already been loaded! at Function.error

These show up only in the console.  They may be harmless or they may be a sign that something isn't working properly or has the potential to not work properly.  In any case, they are distracting, particularly when trying to debug other JavaScript problems.

dlebauer:
fixed in #698

PR #633, opened 2019-01-15
Remove gunload, was deprecated and removed #551

robkooper:
simple fix to prevent gunload message in console #551


dlebauer:
Is this still relevant, should it be merged?

gsrohde:
@dlebauer Yes.  If you open the Developer Tools console, you will see you get a JavaScript error from the `GUnload()` call every time you switch pages.

PR #645, opened 2019-04-17
Create new config file for UA BETY

KristinaRiemer:
I don't know where to put figure files? Or how to set `schema_spy_settings`. @dlebauer 

Closes danforthcenter/sentinel-detection#37

### DIFF
--- a/config/ua-application.yml
+++ b/config/ua-application.yml
@@ -1,0 +1,88 @@
+# To use this file for a University of Arizona BETYDb Web site:
+# 1. Copy this file to config/application.yml
+# 2. Replace the dummy value of rest_auth_site_key with the key appropriate to your site
+# 3. Make any other tweaks you deem appropriate
+# 4. Save the file and restart your BETYdb server
+
+# Contact Information
+admin_phone: "(520) 621-4381"
+admin_email: "dlebauer@email.arizona.edu"
+
+# Override this with a secret key to run a secure site:
+rest_auth_site_key: 'your site's secret key'
+
+schema_spy_settings:
+    java_executable: /usr/java/jre1.8.0_91/bin/java
+    postgresql_driver_jar_file: lib/tasks/jar/postgresql-9.4.1208.jre7.jar
+    settings_for_customized_documentation:
+        schema_spy_jar_file: lib/tasks/jar/schemaSpy.jar
+        output_directory: .
+        remove_root_dir_files: true
+
+organization:
+  name: "University of Arizona"
+  url: #{root_path}
+  logo_file: "ua-logo.png"
+
+site_identification_markup: "<strong>BETY</strong>db <small><strong>Biofuel Ecophysiological Traits and Yields </strong>Database</small>"
+
+# Footer
+
+footer_background_image_file: "setaria.jpg"
+
+citation_license_copyright_markup:
+
+  '<p>LeBauer, David, Rob Kooper, Patrick Mulrooney, Scott Rohde, Dan Wang, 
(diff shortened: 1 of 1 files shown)


KristinaRiemer:
Hey @gsrohde, I've started the config file for changing the appearance of the UA version of the BETYdb website but I'm not sure what all the parts do and am looking for feedback on them if you have time. Thanks!

gsrohde:
@KristinaRiemer 

Kristina:

Unfortunately, I have not yet documented this very thoroughly in the BETYdb documentation files, but there is _some_ information there and there is also some information in issue https://github.com/PecanProject/bety/pull/447.

I suggest that you get a copy of BETYdb running on your own work station if you haven't yet done so so that you can easily test out and tweak your customization file.


Regarding the information in issue #447, there are a few notable changes: Image files no longer go in `/public/images`.  Rather, they should go under `/app/assets/images`.  More specifically, I recommend putting them in the directory `/app/assets/images/custom`.  Then, in your config file, you should refer to them by their paths relative to `/app/assets/images`, for example `custom/my_image_file.jpg`.  (Also note that the `pecandev` server mentioned in that issue is no longer up and running, so again, it's a good idea to deploy on your own work station to test things out.)

There is information about setting the value for `rest_auth_site_key` in Step 9 of the page at https://pecan.gitbook.io/betydb-documentation/installing_betydb/installing-betydb-web-application.  Note that the sample value you have in `ua-application.yml` is invalid.  The reason is that you've single-quoted a value that has an internal single quote (`'your site's secret key'`).  This makes the whole YAML file invalid so that the Rails app won't start.  A solution is to use double quotes to quote the string value.  (Omitting the enclosing quotes altogether also works.)  But of course what you should really do is generate a truly secure key.  Note that you can test that a YAML file is syntactically correct by using the online YAML parser at https://yaml-online-parser.appspot.com/.  You may not want to paste in the "secret" value you should be using for `rest_auth_site_key` though.

Note that if you have been running the site prior to this without any custom configuration file, the passwords for any existing users will become invalid once the value of `rest_auth_site_key` changes from the default value `'thisisnotasecret'` given in `defaults.yml` to whatever value you specify in `application.yml` to override it.  One way to fix this is to create (or recreate) an administrative account following instructions here: https://pecan.gitbook.io/betydb-documentation/installing_betydb/installing-betydb-web-application#step-25-create-a-betydb-administrative-account.  Then log in as this user and reset the passwords for all other users, inform them that their passwords have been reset, and advise them to change their passwords again to something known only to themselves.  You should also do step 26 (https://pecan.gitbook.io/betydb-documentation/installing_betydb/installing-betydb-web-application#step-26-re-set-the-guest-user-account-password), or if there is already a Guest User account, go in and reset its password to `guestuser`.

There is information about setting up SchemaSpy here: https://pecan.gitbook.io/betydb-documentation/installing_betydb/installing-betydb-web-application#step-28-ensure-the-schemaspy-documentation-can-be-generated.  This can always be done later if you prefer—all other parts of the site will work fine without it.

By the way, I don't recommend checking config files and associated image files into the Git repository (which renders this whole pull request moot except as a means of communication).  These are files pertaining to a particular instance of BETYdb—the one at UA—and they therefore don't belong in the code that every user of BETYdb checks out.  Moreover, you certainly shouldn't check in a file containing the value of `rest_auth_site_key` since then everyone in the world will have access to it.  (You may want to keep your own backups of these files though—both the YAML file and any referred-to image files.)

I hope this answers most of your questions.  Let me know if you have further questions or run into difficulties.

dlebauer:
This should be included in a repository with other files required for local deployment at UA. I will close this and create a new issue for @Chris-Schnaufer

Issue #652, opened 2019-06-26
Choose a better list of values for the soil choices on the new site and site editing pages.


gsrohde:
Right now the list of choices is `sandloamy, sandsandy, loamsilt, loamloamsandy, clay, loamsilty, clay, loamclay, loamsandy, claysilty, clayclaypeat`.  Note, among other things, that `clay` is listed twice.  This has been the list since April 4, 2013.  

Prior to that, the list was `sand, loamy sand, sandy loam, silt loam, loam, sandy clay loam, silty clay loam, clay loam, sandy clay, silty clay, clay, peat`.  This seems like a more reasonable list.  I don't know the reason for the change.  This list matches the list laid out in issue #246, which talks about possible constraints on this value. (Right now there are none other than that it be of type `character varying(255)` and not null (but it can be blank).)  I don't remember what discussion, if any, gave rise to the list in issue #246, but perhaps we considered restoring that as the list of possible choices provided in the BETYdb Rails app in addition to establishing a database constraint.

The reason for prioritizing this issue is that I would like to establish a set of ideal soil value choices before adding new sites to the TERRA-MEPP database in connection with https://github.com/terra-mepp/data/issues/12.

@dlebauer Could you please take a look at this and comment?

dlebauer:
We can use the list of soil texture classifications from the USDA soil texture triangle, plus peat, bedrock, and 'other':

![image](https://user-images.githubusercontent.com/464871/61312447-46c4e380-a7ad-11e9-8e0a-a31f05611a36.png)

gsrohde:
Completed by release BETYdb 5.2.1.  Closing.

PR #655, opened 2019-07-18
Modify site soil choices

gsrohde:
I used the 12 USDA types (in alphabetical order) followed by `peat`, `bedrock`, and `other` (in that order).  That is, this is the order the soil classes will appear in the `Soil` dropdown in the forms on the Site creation and modification pages.  The order could be changed if desired.

This change set addresses issue #652 .


gsrohde:
Update of soil type choices.

gsrohde:
Update of soil type choices.

gsrohde:
Update of soil type choices.

PR #657, opened 2019-08-05
update documentation links

dlebauer:
reflecting migration to bookdown


gsrohde:
@dlebauer Looks good except you should probably update the reference in `script/db_maintenance/README` as well.

dlebauer:
@gsrohde thanks - I've updated that link as well.

PR #658, opened 2019-08-05
HOT FIX for Bulk Upload method specification problem

gsrohde:
This fixes the bug mentioned in this issue: https://github.com/az-digitalag/organization/issues/12

The bug was introduced by the upgrade to Rails 5.1, which no longer supports "render nothing: true".  (See, e.g., https://stackoverflow.com/questions/34688726/the-nothing-option-is-deprecated-and-will-be-removed-in-rails-5-1.)


robkooper:
This is missing a bump in release number as well as a changelog entry.

robkooper:
Also make sure to pull master back into develop.

PR #660, opened 2019-08-06
Fix method specification in bulk upload

robkooper:
update changelog and travis


gsrohde:
Travis got through the bundle install step okay, but now it's hung up on the next step, adding the postgis extension.

robkooper:
@gsrohde tests are now failing, can you look at this.

PR #661, opened 2019-08-07
update ruby version

robkooper:
Removed ruby 2.2.4 and added latest of 2.3, 2.4, 2.5 and 2.6 versions


robkooper:
Correct, however if I use 2.2.10 it complains stating that 2.2.10 is not new enough since it requires 2.2.2, this is why I now use 2.2.9

PR #665, opened 2019-08-08
In get_insertion_data, change the way interactively_specified_values is set.

gsrohde:
In Ruby >= 2.4, `nil.clone` succeeds.  So we can no longer rely on `@session["global_values"].clone`
throwing an exception when `@session["global_values"]` is `nil` and we must
explicitly check whether `@session` has the key `:global_values`.


gsrohde:
@robkooper Merge this into `release/5.2.1` instead (and update the change log) if you think it appropriate.

PR #678, opened 2019-09-17
set default show_crop_map_links: false

dlebauer:
in defaults.yml

1. this should be off by default; they were only relevant to BETYdb.org
2. Google has deprecated Fusiontables that these maps use; easiest fix is not to show them. No longer a priority mandated by EBI

### DIFF
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -92,7 +92,7 @@ homepage_body:
 # Miscellaneous
 
 ## This affects both the home page and the footer:
-show_crop_map_links: true
+show_crop_map_links: false
 
 # Override this with a secret key to run a secure site:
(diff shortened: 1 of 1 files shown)


dlebauer:
The maps no longer work on google fusion tables

gsrohde:
@dlebauer You seem to be right, although I thought Fusion Tables weren't going away until December 3.

In any case, if we no longer want the maps, a cleaner solution would be to simply remove all the code associated with them rather than simply changing the default.

robkooper:
@dlebauer do you want to update this to remove the maps all together?

dlebauer:
yep - we should remove the maps all together - google fusion tables and associated maps have been deprecated

PR #681, opened 2019-12-06
security and zenodo fixes 

dlebauer:
udpated a few dependency versions
fixed .zenodo.json causing broken archiving


robkooper:
Can you update the Changelog and a new version number, we need to tag it next with a release to have zenodo build.

PR #687, opened 2020-01-14
Make defaults generic

dlebauer:
It is confusing enough that there are so many instances of betydb. It does not make sense to have the default settings look like the database at betydb.org:

image

One additional step would be to add a table of all of the databases that people might be looking for.

### DIFF
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -4,16 +4,15 @@
 
 
 organization:
-  name: "Energy Biosciences Institute"
+  name: "Organization Name Here"
   url: #{root_path}
-  logo_file: "logo-ebi.png"
+#  logo_file: "logo-file-here.png"
 
-site_identification_markup: "<strong>BETY</strong>db <small><strong>Biofuel Ecophysiological Traits and Yields </strong>Database</small>"
-
+site_identification_markup: "<strong>GENERIC</strong>db <small><strong>Generic PEcAn, Ecological, and/or Agronomic </strong>Database</small>"
 
 # Footer
 
-footer_background_image_file: "logo-ebi-footer.jpg"
+#footer_background_image_file: "logo-footer.jpg"
 
 ## Contact Information
 admin_phone: "(000) 000-0000"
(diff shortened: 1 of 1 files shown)


robkooper:
@dlebauer can you add an image of what it looks like now with the changes.

dlebauer:
@robkooper working on it ... once I run 

```
docker-compose -p bety up -d postgres
docker run --rm --network bety_bety pecan/bety initialize
docker-compose -p bety up -d
```

Should the interface be at localhost:8000/bety?

robkooper:
Yes, should be.  Make sure that port is exposed in bh the docker-compose file

dlebauer:
@robkooper I have the same docker-compose.yml as the repo ... looks like the port should be exposed

```
services:
  # BETY rails frontend to the database
  bety:
    image: pecan/bety:${BRANCH:-latest}
    networks:
      - bety
    ports:
      - 8000
    environment:
      - UNICORN_WORKER_PROCESSES=1
      - SECRET_KEY_BASE=thisissomereallllllylongsecretkeyandshouldbelongerthanthis
    depends_on:
      - postgres
    restart: unless-stopped
```

robkooper:
correct at that point it is exposed, not as port 8000 though. You will need to do a `docker ps` to see the port, or easier `docker port bety 8000` You will see something like: `0.0.0.0:32769` meaning you need to connect to http://localhost:32769.

config/defaults.yml
```diff
@@ -4,79 +4,77 @@
...
   covered by this license but may be available for use with consent.</p>
 
-  <p>Copyright © 2010-2014 Energy Biosciences Institute</p>'
+'
 
 sponsors:
-    - URL: http://illinois.edu
-      title: "University of Illinois at Urbana-Champaign"
-      text: "University of Illinois at Urbana-Champaign"
+    - URL: http://example.com
```
robkooper, line 42: how about we replace this with icons and names of all universities involved that contributed to BETY
- Illinois
- BU
- AU

PR #690, opened 2020-03-30
use public. prefix in functions

robkooper:
when doing a restore some functions would fail to find other functions or tables. Prefixing these with public. fixes problem.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ For more information about this file see also [Keep a Changelog](http://keepacha
 - update doi links http://dx.doi.org --> https://doi.org
 - #672 : Added activemodel-serializers-xml Gem to restore functionality of "original" API XML endpoints.
 - #674 : Upgraded comma Gem to restore CSV file downloads.
+- fixed functions to explicitly add public. to function call and table reference. (this will require a migration)
```
Chris-Schnaufer: Perhaps add link on how to perform migration?

robkooper: see https://pecanproject.github.io/bety-documentation/technical/updating-betydb-when-new-versions-are-released.html and 7.2

This will be added to the changelog as well.

PR #691, opened 2020-03-30
docker image with databae

robkooper:
creates docker image pecan/betydb that contains full dump of database and script to quickly restore.

This requires pull request #690


initdb/README.md
```diff
@@ -0,0 +1,27 @@
+# Dump DB to Docker
+
+Running the `build.sh` command will create a docker image, `pecan/betydb`, which contains
```
Chris-Schnaufer: I find the default name shown here confusing in regards to the real database. Perhaps 'pecan/betydb_backup' could be clearer?

robkooper: it can be a backup, but is not necessarily, I see it used for distributing a database to initialize as well.

@dlebauer do you have any ideas of what we can call it? Maybe `pecan/bety_database`

dlebauer: @robkooper I can see why pecan/betydb is not specific and how its not really a backup. Maybe betydb_public? This gets back to the fact that 'betydb' is confusing because it has many meanings (betydb.org, a software stack, a schema, multiple networks of databases ...).

infotroph: `pecan/minimal_db`?

robkooper: maybe pecan/db?

dlebauer: sounds good!

PR #693, opened 2020-04-09
Switch from travis to GitHub Actions

robkooper:
Uses github actions to build BETY


robkooper:
before merging squash all commits.

PR #694, opened 2020-04-09
Use github actions instead of travis

robkooper:
This moves the building from travis to github actions. Builds will be done on pull request and push to master/develop branches as well as something is tagged

When a tag is made, it will push the images to github package registry.


robkooper:
lets make sure we use merge squash.

robkooper:
Not sure about travis, I removed it as a must pass check and think most people would just ignore it. From all the testing I had to do to get it to build I can say that it will fail if there is something wrong :)

robkooper:
As for the release, the real test is to see if it builds when merged into develop.

PR #695, opened 2020-04-09
dumping DB + bug fix

### LINKED ISSUE #696: SECRET_KEY_BASE not used
The variable SECRET_KEY_BASE is actually not used. Setting this variable will not change the encryption key. This is actually set in config/default.yml

robkooper:
this will use github actions to make the nightly db dump
added changes to fix #696 
added handy script to add new users
updated load.bety.sh to allow to fix/set local bety ID.


robkooper:
I will squash this again to minimize the number of commits.

robkooper:
want to add to restore script ability to set starting DB ID as well guest and carya user

robkooper:
@gsrohde if you have some time to look at the issue with SECRET_KEY_BASE and see if I made any errors in ruby (did some testing and it looks ok).

.github/workflows/release.yml
```diff
@@ -24,7 +24,9 @@ jobs:
       - name: Login into registry
         run: |
           echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
-          echo "${{ secrets.DOCKERHUB }}" | docker login -u kooper --password-stdin
+          if [ -n "${{ secrets.DOCKERHUB_USERNAME }}" -a -n "${{ secrets.DOCKERHUB_PASSWORD }}"]; then
```
infotroph, line 27:
```suggestion
          if [ -n "${{ secrets.DOCKERHUB_USERNAME }}" -a -n "${{ secrets.DOCKERHUB_PASSWORD }}" ]; then
```

.github/workflows/release.yml
```diff
@@ -62,6 +64,8 @@ jobs:
           for T in $TAGS; do
             docker tag image $IMAGE_ID:$T
             docker push $IMAGE_ID:$T
-            docker tag image pecan/bety:$T
-            docker push pecan/bety:$T
+            if [ -n "${{ secrets.DOCKERHUB_USERNAME }}" -a -n "${{ secrets.DOCKERHUB_PASSWORD }}"]; then
```
infotroph, line 67:
```suggestion
            if [ -n "${{ secrets.DOCKERHUB_USERNAME }}" -a -n "${{ secrets.DOCKERHUB_PASSWORD }}" ]; then
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ In this case you can simply pull down the version you want, and run `docker-comp
 ## [Unreleased]
 
 ### Fixes
+- #696 : actually now uses environment variable SECRET_KEY_BASE. If this was set, unset otherwise you can not login to BETY.
```
infotroph, line 23:
```suggestion
#696 : breaking change: Now actually uses environment variable SECRET_KEY_BASE, which was silently ignored before. If this was set on your system, unset it or you will not be able to log in to BETY.
```

Maybe too long now? But I think the current wording would make it easy for me to miss that this requires user action if I were reading the changelog without having seen the PR description.

robkooper: My goal is to put this in **BOLD** in the release notes.

PR #702, opened 2020-04-28
remove static content

robkooper:
I don't think has been used in ages.


gsrohde:
I wouldn't swear to it, but I think you are correct that most of this stuff isn't used anymore.  But, last I knew, we still use 404.html when users try to go to a non-existent page.

robkooper:
yup, that is why the build failed.

PR #740, opened 2022-11-14
Splitting input files in load.bety.sh

Chris-Schnaufer:
Large files on underpowered machines can cause the to appear to database to hang through locking conflicts. One cause is that loading large files in one go uses significant resources, slowing the loading process down. Breaking large source files apart and loading them in a loop resolves this issue

### DIFF
--- a/script/load.bety.sh
+++ b/script/load.bety.sh
@@ -405,8 +405,18 @@ trap '
 # 3) load new data
 # 4) set last inserted item in my range
 # 5) enable constraints on this table
+PREV_SPLIT=""
 for T in ${EMPTY_TABLES} ${CLEAN_TABLES} ${CHECK_TABLES} ${MANY_TABLES}; do
-  # start
+  # Start
+  if [ "${PREV_SPLIT}" != "" ]; then
+    if [ "${QUIET}" != "YES" ]; then
+      echo "Cleaning up previous split: ${PREV_SPLIT}"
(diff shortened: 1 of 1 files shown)


dlebauer (approved):

dlebauer:
@kooper please review

script/load.bety.sh
```diff
@@ -404,7 +428,27 @@ for T in ${EMPTY_TABLES} ${CLEAN_TABLES} ${CHECK_TABLES} ${MANY_TABLES}; do
     echo "SELECT COUNT(*) FROM ${T};" >&3 && read START <&4
     if [[ "${EMPTY}" == "NO" || ${EMPTY_TABLES} == *"$T"* ]]; then
       if [ -f "${DUMPDIR}/${T}.csv" ]; then
-        echo "\COPY ${T} FROM '${DUMPDIR}/${T}.csv' WITH (DELIMITER '	',  NULL '\\N', ESCAPE '\\', FORMAT CSV, ENCODING 'UTF-8')" >&3
+	T_SIZE=`cat ${DUMPDIR}/${T}.csv | wc -l`
+	if [ "${T_SIZE}" -gt "${SPLIT_FILE_MAX}" ]; then
```
robkooper, line 432: can we have SPLIT_FILE_MAX==0 means use the old algorithm, and for that to be the default value?

PR #751, opened 2023-03-28
[WIP] Add ruby 2.7 to build

dlebauer:
2.5 and 2.6 have hit EOL. 
2.7 build is successful. This can be merged 

But, should we remove 2.5 and 2.6 from the build since they have reached EOL? 

And update the docker image to use 2.7?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
         ruby: 
           - 2.5
           - 2.6
-          # - 2.7 is not supported, results in error
+          - 2.7 is not supported, results in error
     services:
       postgres:
         image: mdillon/postgis:9.5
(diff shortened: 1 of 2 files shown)


robkooper:
can you update it to use checkout@v3 as well to get rid of the nodejs warnings.?

dlebauer:
@robkooper sure 
- and is there any need to keep deprecated ruby versions 2.5 and 2.6?
- I'll go ahead and test with non-deprecated postgis in a separate branch

robkooper:
I think we can get rid of the other version of ruby, we do need to make sure that the dockerfile uses ruby 2.7

dlebauer:
Closing this PR: Ruby 2.7 updates have been added to https://github.com/PecanProject/bety/pull/753

PR #758, opened 2024-07-13
switch to ruby 3.1

robkooper:
replaces #738


dlebauer (approved):

robkooper:
Looking more, this will be a significant lift:

[Rails 6.0.2](https://github.com/rails/rails/releases/tag/v6.0.3) is the earliest Rails version which is compatible with Ruby 3.0.

